Issue #2, opened 2017-10-13
Migrate tests to use pytest


jonc125:
Rather than the legacy scons-based framework

jonc125:
This was partially done in #6, in that pytest is now used to run the tests, but there are a few hacks still present to cope with assumptions from Chaste's framework, and in particular `TestPythonReproducibilty` won't work (even with #5 done).

PR #6, opened 2018-12-04
Make fc a standard Python package

### LINKED ISSUE #1: Restructure file hierarchy & make pip installable
With this being its own repo for the Python backend, we need to move things into a more logical structure, and make it into a 'proper' Python package.

- [ ] Figure out the best structure. Need to consider:
    - Nicely integrating the Cython models, which will be generated on-the-fly
    - Testing (#2)
- [ ] Implement it!
- [ ] Installation of dependencies. We require sundials, which can be obtained with `conda install sundials -c conda-forge`
- [ ] Move setup.py to top level & extend
- [ ] Documentation #4

jonc125:
Fixes #1, and starts on #4 and #2.


jonc125:
And we just had a passing build!

Ready to review now @MichaelClerx :)

fc/language/expressions/abstract.py
```diff
@@ -155,17 +123,17 @@ def GetDefiningEnvironments(self, env):
...
+
         TODO: Handle local name conflicts!
         """
         try:
             return self._definingEnvs
         except AttributeError:
             self._rootDefiningEnv = env  # For paranoia checking that the cache is valid
             d = self._definingEnvs = {}
-            l = self._usedVarLocalNames = []
+            l = self._usedVarLocalNames = []  # noqa: E741
```
MichaelClerx: Should we disable this one project-wide?

fc/language/expressions/array.py
```diff
@@ -71,20 +40,20 @@ def Interpret(self, env):
...
             raise NotImplementedError
-        
+
         result = ne.evaluate(compiled_gen_expr, local_dict, env.unwrappedBindings)
         return result
-        
+
     def DevelopResultWithInterpret(self, range_dims, range_name, ranges, env, num_gaps, dims):
-        #print 'Comprehension', self.location
+        # print 'Comprehension', self.location
```
MichaelClerx: This is another flake8 (or actually pep?) rule that I'm happy to disable

fc/language/expressions/array.py
```diff
@@ -420,14 +401,14 @@ def Interpret(self, env):
...
-        
+
         total_so_far = initial
-       
-        for i,dim in enumerate(shape):
+
+        for i, dim in enumerate(shape):
             if i == dimension:
-                dim_ranges.append(range(1))
+                dim_ranges.append(list(range(1)))
```
MichaelClerx: [0] ?

fc/language/expressions/array.py
```diff
@@ -562,57 +551,61 @@ def Interpret(self, env):
...
             end = num_entries
             move = 1
-        
-        # The next_index array keeps track of how far along dimension dim_val we should put the next kept entry at each location
+
+        # The next_index array keeps track of how far along dimension dim_val we
+        # should put the next kept entry at each location
         shape[dim_val] = 1
-        next_index = np.zeros(shape)
+        next_index = np.zeros(shape, dtype=np.uint32)
```
MichaelClerx: Any reason for this particular type?

fc/language/expressions/array.py
```diff
@@ -71,20 +40,20 @@ def Interpret(self, env):
     def DevelopResultWithCompile(self, range_name, ranges, compiled_gen_expr, env):
```
MichaelClerx: Are we sticking with CamelCase for this one? (Actually it's PascalCase isn't it?)

fc/language/statements.py
```diff
@@ -53,7 +23,7 @@ def __init__(self, names, rhs, optional=False):
     def Evaluate(self, env):
         try:
             results = self.rhs.Evaluate(env)
-        except:
+        except BaseException:
```
MichaelClerx: Why are we catching system-exciting exceptions here? (Rather than catching Exception)

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1481,43 +1498,50 @@ class CompactSyntaxParser(object):
...
-                                        ('^', 2, p.opAssoc.LEFT, Actions.Operator),
-                                        ('-', 1, p.opAssoc.RIGHT, lambda *args: Actions.Operator(*args, rightAssoc=True)),
-                                        (p.oneOf('* /'), 2, p.opAssoc.LEFT, Actions.Operator),
-                                        (p.oneOf('+ -'), 2, p.opAssoc.LEFT, Actions.Operator),
-                                        (p.Keyword('not'), 1, p.opAssoc.RIGHT, lambda *args: Actions.Operator(*args, rightAssoc=True)),
-                                        (p.oneOf('== != <= >= < >'), 2, p.opAssoc.LEFT, Actions.Operator),
-                                        (p.oneOf('&& ||'), 2, p.opAssoc.LEFT, Actions.Operator)
-                                       ])
-    
+    expr <<= p.infixNotation(atom, [(accessor, 1, p.opAssoc.LEFT, Actions.Accessor),
```
MichaelClerx: Yikes

MichaelClerx: Is there some way to write this that makes it more obvious what's going on?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -109,18 +70,20 @@ class Actions(object):
 
     class BaseAction(object):
         """Base parse action.
```
MichaelClerx: What's a "base parse action"?

fc/utility/test_support.py
```diff
@@ -39,24 +8,27 @@
...
-    
+
     Numbering starts from 1.  If this process is not one started by the multiprocessing module, 0 is returned.
     """
     import multiprocessing
     name = multiprocessing.current_process().name
     try:
         num = int(name.split('-')[-1])
-    except:
+    except BaseException:
```
MichaelClerx: As above

test/python/TestAlgebraicModels.py
```diff
@@ -1,52 +1,25 @@
...
     """Test behaviour on models with no ODEs; based on TestClamping.hpp"""
-    def TestClampingToInitialValue(self):
-        proto = fc.Protocol('projects/FunctionalCuration/test/protocols/test_clamping1.txt')
+
+    @pytest.mark.xfail(strict=True, reason='no pycml replacement yet')
+    def testClampingToInitialValue(self):
+        proto = fc.Protocol('test/protocols/test_clamping1.txt')
         proto.SetOutputFolder('Py_TestAlgebraicModels_TestClampingToInitialValue')
-        proto.SetModel('projects/FunctionalCuration/cellml/beeler_reuter_model_1977.cellml')
+        proto.SetModel('cellml/beeler_reuter_model_1977.cellml')
```
MichaelClerx: os.path.join

.gitignore
```diff
@@ -4,18 +4,44 @@
...
-/src/python/fc/sundials/solver.c
-/src/python/fc/sundials/solver.so
-/src/python/fc/sundials/sundials.so
-
-# Eclipse stuff
-/.pydevproject
-/.settings
+*.py[cod]
 
 # CellML project, in case we've synced from a build machine
```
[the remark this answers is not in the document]
jonc125: That's left over from Chaste days. Once we test with actual CellML files we'll either add as a submodule or just get Travis to clone both repos.

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+cache: pip
+python:
+  - "3.6"
+addons:
+  apt:
+    packages:
+      - libsundials-serial-dev
+install:
+  # Needed to run setup.py
+  - pip install cython numpy
```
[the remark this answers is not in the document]
jonc125: No, unfortunately. I did try putting them in, but setuptools can't get as far as realising it needs to install them without first trying to import them...

fc/language/expressions/abstract.py
jonc125: Not sure. It's generally a sensible warning, except for very small bits of code like this where there's no chance of confusion!

fc/language/expressions/array.py
jonc125: Most editors will comment with `# ` for you if you use the 'comment this block' shortcut...

fc/language/expressions/array.py
jonc125: It needs to be an integer type so we can use it to index numpy arrays, and it doesn't need to be negative, so I went for 32-bit unsigned

fc/language/expressions/array.py
jonc125: I didn't think it was worth going through and changing everything away from the Chaste convention!

test/python/TestAlgebraicModels.py
jonc125: ?

fc/language/statements.py
jonc125: Because that's what 2to3 did by default and I didn't think about it. I'll change these.

fc/parsing/CompactSyntaxParser.py
jonc125: Not easily! It's not too bad if you're familiar with the pyparsing approach to parsing infix notation. Basically, as the comment 2 lines up says, this is the main grammar for parsing expressions in the post-processing language. So it has to be fairly complex.

fc/language/expressions/array.py
MichaelClerx: I'd rather just type a #  :D

fc/language/expressions/array.py
MichaelClerx: Ok. Nothing wrong with that! Only it makes it look like there's going to be some sort of interaction with low-level code at one point, or maybe packing into a file or such. Would there be any downside to just writing dtype=int ?

test/python/TestAlgebraicModels.py
MichaelClerx: Is it not generally discouraged to write paths with slashes, and to use os.path.join instead?

(Although the only practical case where it matters is outdated macs)

jonc125: I have a vague recollection that some parts of Python at least are clever enough to translate paths written directly as strings. In any case, we're not going to be running this on Windows, and writing the paths directly looks tidier!

PR #28, opened 2019-03-22
Dynamic code generation test

MichaelClerx:
Added a test for very basic automatic code generation.

At the moment this requires you to manually run `requirements/weblab_cg.sh` to install the correct branch of `cg` as a dependency


fc/environment.py
```diff
@@ -2,7 +2,7 @@
 import numpy as np
```
jonc125: I'm not sure about the rationale for moving this file (unlike moving protocol.py which I agree with)?

Similarly the other utility modules below.

fc/parsing/CompactSyntaxParser.py
```diff
@@ -996,7 +1015,7 @@ def _expr(self):
             args.extend([output_names, optional_flags])
             model = Model.NestedProtocol(*args)
             result = Simulations.OneStep(0)
-            result.SetModel(model)
+            result.set_model(model)
```
jonc125: I'd rather not do this kind of renaming piecemeal, as it leads to an inconsistent API. Rather we should do it in one (huge!) PR for the entire project.

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1807,16 +1838,21 @@ def __del__(self):
         sys.setrecursionlimit(self._original_stack_limit)
 
     def _Try(self, callable, *args, **kwargs):
-        """Try calling the given parse command, increasing the stack depth limit if needed."""
+        """
+        Try calling the given parse command, increasing the stack depth limit
```
jonc125: What max line width are you working to? I think other parts of Web Lab code are using 120.

(really need to set up flake8 config for everything!)

fc/protocol.py
```diff
@@ -103,21 +100,37 @@ def __init__(self, protoFile, indentLevel=0):
...
+
+        # A mapping of namespace names to uris, for namespaces used in the
+        # protocol
+        self.ns_map = {}
+
         # Parse the protocol file and fill in the structures declared above
         start = time.time()
+
         import fc.parsing.CompactSyntaxParser as CSP
+        CSP.DoXmlImports()
```
jonc125: Why is this needed?

fc/protocol.py
```diff
@@ -135,14 +148,28 @@ def __init__(self, protoFile, indentLevel=0):
                 self.postProcessing.extend(imported_proto.postProcessing)
                 self.outputs.extend(imported_proto.outputs)
                 self.plots.extend(imported_proto.plots)
+                self.model_interface.extend(imported_proto.model_interface)
+                for prefix, uri in imported_proto.ns_map.items():
```
jonc125: We probably ought to have some error thrown if a prefix ends up getting bound to a different URI here!

fc/environment.py
MichaelClerx: An environment seems like a core concept in FC to me, no? I'd think a utility was something like a CSV parser or some other stand-alone feature that's not very FC-specific but needed anyway

fc/parsing/CompactSyntaxParser.py
MichaelClerx: OK, will revert!

fc/environment.py
jonc125: On that basis, I think everything in `utility` should be moved into the top level folder!

fc/parsing/CompactSyntaxParser.py
MichaelClerx: I've left it at the standard PEP8 width of 79.
It's the standard, but I personally really like that length as well as it means I can fit an editor and a terminal side by side on my laptop screen.

fc/protocol.py
MichaelClerx: I don't know! It doesn't work if that's never called. I think we discussed it was something about the way the compact syntax parser was designed at the moment that we were hoping to eventually change?

fc/environment.py
MichaelClerx: ok!

fc/parsing/CompactSyntaxParser.py
MichaelClerx: (This change was more because of the `"""Starting like this` which I've always found very ugly. But that's a personal preference sorry!)

fc/protocol.py
MichaelClerx: Good point! How do we raise errors? Do we have some kind of special Exception type that gets converted to user-friendly messages?

MichaelClerx: Good point. Have added a test for this

PR #38, opened 2019-05-07
Remove xml output code from CompactSyntaxParser

MichaelClerx:
Part of #31

### DIFF
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -513,13 +513,9 @@ class Actions(object):
     ######################################################################
 
     class SetTimeUnits(BaseAction):
-        #def _xml(self):
-        #    return P.setIndependentVariableUnits(**self.TransferAttrs('units'))
         pass
 
     class InputVariable(BaseGroupAction):
-        #def _xml(self):
-        #    return P.specifyInputVariable(**self.TransferAttrs('name', 'units', 'initial_value'))
         pass
 
     class OutputVariable(BaseGroupAction):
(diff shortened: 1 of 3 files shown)


jonc125:
We will be missing some coverage at this level without the `ParseFile` tests, yes, as they check real protocols can be parsed successfully. However, tests elsewhere in the codebase parse and run real protocols, so should ensure everything is covered somewhere. So probably OK!

MichaelClerx:
OK. Ready for review then I guess!

fc/parsing/CompactSyntaxParser.py
```diff
@@ -46,7 +46,7 @@ def DoXmlImports():
         globals()[name] = local_defs[name]
 
 
-if __name__ == '__main__' or getattr(sys, '_fc_csp_no_pyimpl', False):
+if getattr(sys, '_fc_csp_no_pyimpl', False):
```
jonc125: This shouldn't be needed if we're removing the XML generation.

fc/parsing/CompactSyntaxParser.py
```diff
@@ -726,16 +513,15 @@ def _expr(self):
     ######################################################################
 
     class SetTimeUnits(BaseAction):
-        def _xml(self):
-            return P.setIndependentVariableUnits(**self.TransferAttrs('units'))
+        pass
```
jonc125: With all these model interface sections, it would be worth storing the useful properties as member variables in some fashion, or at least documenting what is available, to assist with replacing the protocol parts of pycml later.

jonc125: For instance, the `TransferAttrs` function is only needed by XML generation. But it'd probably be handy to keep a renamed version that does the extraction as a dict, and save that dict as a member var.

fc/parsing/CompactSyntaxParser.py
```diff
@@ -960,14 +648,7 @@ def _expr(self):
...
     class OneStepSimulation(BaseGroupAction):
-        def _xml(self):
-            attrs = {}
-            args = []
-            if 'step' in self.tokens:
-                attrs['step'] = str(self.tokens['step'][0])
-            if 'modifiers' in self.tokens:
-                args.append(self.tokens['modifiers'][0].xml())
-            return P.oneStep(*args, **attrs)
+        pass
```
jonc125: Again here we should document what the useful fields are. The one-step simulation class isn't (yet?) implemented in the Python version of FC.

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1115,39 +746,15 @@ def GetValue(self, token, negate=False):
                         result = '-' + result
             return result
 
-        def _xml(self):
```
jonc125: Again, document what's useful. We'll need to generate pint unit definitions from these, probably.

test/TestCompactSyntaxParser.py
```diff
@@ -182,41 +105,39 @@ def testParsingComments(self):
...
-        self.assertParses(csp.expr, 'var?', [['var']], ('ci:var', trace))
-        self.assertParses(csp.expr, '(1 + a)?', [[['1', '+', 'a']]], ('apply', trace, ['plus', 'cn:1', 'ci:a']))
-        self.assertParses(csp.expr, '1 + a?', [['1', '+', ['a']]], ('apply', ['plus', 'cn:1', ('ci:a', trace)]))
+        self.assertParses(csp.expr, '1?', [['1']])
+        self.assertParses(csp.expr, 'var?', [['var']])
+        self.assertParses(csp.expr, '(1 + a)?', [[['1', '+', 'a']]])
+        self.assertParses(csp.expr, '1 + a?', [['1', '+', ['a']]])
 
     def testParsingMultiLineExpressions(self):
+
```
jonc125: Why a blank line?

test/TestCompactSyntaxParser.py
```diff
@@ -182,41 +105,39 @@ def testParsingComments(self):
...
+        self.assertParses(csp.expr, '0 || 1 && 1', [['0', '||', '1', '&&', '1']])
         self.assertParses(csp.expr, 'A + B || C * D', [[['A', '+', 'B'], '||', ['C', '*', 'D']]])
-        self.assertParses(csp.expr, 'if 1 then 2 else 3', [['1', '2', '3']],
-                          ('piecewise', [('piece', ['cn', 'cn']), ('otherwise', ['cn'])]))
+        self.assertParses(csp.expr, 'if 1 then 2 else 3', [['1', '2', '3']])
         self.assertParses(csp.expr, 'if 1 < 2 then 3 + 4 else 5 * 6',
                           [[['1', '<', '2'], ['3', '+', '4'], ['5', '*', '6']]])
 
     def testParsingTrace(self):
+
```
jonc125: Why a blank line?

test/TestCompactSyntaxParser.py
```diff
@@ -227,16 +148,13 @@ def testParsingMultiLineExpressions(self):
         self.assertParses(csp.expr, '(1 + 2)\n * 3', [[['1', '+', '2'], '*', '3']])
 
     def testParsingSimpleAssignments(self):
-        self.assertParses(csp.simpleAssign, 'var = value', [['var', 'value']],
-                          ('apply', ['eq', 'ci:var', 'ci:value']))
+
```
jonc125: Why a blank line?

jonc125: I'll stop mentioning these below!

PR #45, opened 2019-07-18
Commented out line-profiler

### LINKED ISSUE #43: pip install fails due to `line-profiler`
`pip install -e .[dev,test]`

Somewhere down the line, something pulls in `line-profiler`, which fails to install:

> Note: As of version 2.1.2, pip install line_profiler does not work. Please install as follows until it is fixed in the next release:

https://github.com/rkern/line_profiler

MichaelClerx:
Closes #43.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,7 +82,7 @@ setup(
         'matplotlib',
         'numexpr',
         'numpy',
-        'pyparsing',
+        'pyparsing<2.4',
         'scipy',
         'tables',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -89,7 +89,7 @@
     ],
     extras_require={
         'dev': [
-            'line_profiler',
+            #'line_profiler',
```
jonc125: There's a flake8 error:
`./setup.py:92:13: E265 block comment should start with '# '`

Issue #25, opened 2019-02-19
Fix coverage testing


MichaelClerx:

MichaelClerx:
@jonc125 I'm trying to fix it but unable to make it work using pytest-cov.
Can we just switch to ``coverage -m pytest`` instead?

MichaelClerx:
Oh wait think I found it: It runs code from the venv, not the local code

PR #53, opened 2019-09-11
Fix coverage reporting

MichaelClerx:
See #25

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,9 @@
+[run]
+omit =
+source = fc
+branch = True
+
+[report]
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,6 @@ install:
   - flake8 --version
 script:
   - flake8
-  - pytest --cov test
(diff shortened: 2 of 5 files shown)


MichaelClerx:
Uhm... Why is the coverage config file stored inside the python module there??

jonc125:
Ah, though having approved, I've just realised that the coverage is reported at 0% again...

MichaelClerx:
OK. It worked locally because I was still using the old -e install. Now copying the cellmlmanip configuration but to no avail

MichaelClerx:
I'm happy for `tests` to be a module anyway, so that the tests can e.g. share code

MichaelClerx:
OK. Now for some reason the Cython classes weren't found. Re-running and hoping it works this time round :-/

MichaelClerx:
This makes me want to just install with `-e` again @jonc125 ! Our users can't pip install anyway, because they need numpy and cython pre-installed

MichaelClerx:
Weird error this! Can't recreate it locally, following the same steps as travis

MichaelClerx:
Any ideas @jonc125 ? I don't have much experience with Cython

PR #54, opened 2019-09-13
Try minimal coverage fixes in a new branch

### LINKED ISSUE #25: Fix coverage testing

jonc125:
An alternative to #53 that attempts to fix #25.


MichaelClerx:
Wondering what causes this!
Why would it not be able to find the files created by pip install? Are they not included in the package maybe (since they're binaries instead of .py files)?

jonc125:
I wish I knew! It certainly appears to be including the .so and .c files in the package and installing them to site-packages...

jonc125:
Anyway, shall we merge this and close #53? Maybe open an issue for later to figure out the root cause?

Issue #19, opened 2019-02-18
Refactor to use python style names: ClassName.method_name


MichaelClerx:
Similarly, variables all `like_this` never `likeThis` or `LikeThis`

MichaelClerx:
Major part of this is tackled by #35 

Remaining names to look at (as identified by some regexes):

- [ ] AddLoc
- [ ] Adjacent
- [ ] Delegate
- [ ] DelegateSymbol
- [ ] DelimitedMultiList
- [ ] DisableDebug
- [ ] DisplayLoc
- [ ] EnableDebug
- [ ] ExceptionDebugAction
- [ ] GetAttributeDict
- [ ] GetChildrenExpr
- [ ] GetChildrenXml
- [ ] GetNamedGrammars
- [ ] GetValue
- [ ] MakeKw
- [ ] MakeTestSuite
- [ ] MonkeyPatch
- [ ] N
- [ ] NullRange
- [ ] Operator
- [ ] OperatorOperands
- [ ] OptionalDelimitedList
- [ ] ParseFile
- [ ] RunExperiment
- [ ] SuccessDebugAction
- [ ] Symbol
- [ ] UnIgnore
- [ ] WorkerInit
- [ ] _BindingsDict
- [ ] _FreeVarList
- [ ] _OutputsList
- [ ] _Try
- [ ] _UnwrappedBindingsDict
- [ ] assertAlmostEqual
- [ ] assertDoesNotParse
- [ ] assertEqual
- [ ] assertFalse
- [ ] assertFilesMatch
- [ ] assertIsInstance
- [ ] assertNotEqual
- [ ] assertNotIn
- [ ] assertParses
- [ ] assertRaises
- [ ] assertTrue
- [ ] checkParseResults
- [ ] parseImpl
- [ ] setUpClass
- [ ] tearDownClass
- [ ] testExperimentReproducibility
- [ ] testParsingTrace
- [ ] test_parsing_MathML_funcs
- [ ] test_simple_ODE

MichaelClerx:
And some more (including arguments):

- [ ] A
- [ ] Array
- [ ] False
- [ ] However
- [ ] Library
- [ ] Max
- [ ] Min
- [ ] N
- [ ] None
- [ ] Null
- [ ] OutputFolder
- [ ] ParserElement
- [ ] Plus
- [ ] ProtocolError
- [ ] Similarly
- [ ] Symbol
- [ ] True
- [ ] assignStmt
[...]

PR #58, opened 2019-09-16
Further refactoring for PEP8 names.

MichaelClerx:
See #19

### DIFF
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -921,18 +921,18 @@ class Optional(p.Optional):
     def __init__(self, *args, **kwargs):
         super(Optional, self).__init__(*args, **kwargs)
         self.callPreparse = False
-        self._optionalNotMatched = p.Optional(p.Empty()).default_value
+        self._optionalNotMatched = p.Optional(p.Empty()).defaultValue
 
     def parseImpl(self, instring, loc, doActions=True):
         try:
             loc, tokens = self.expr._parse(instring, loc, doActions, callPreParse=True)
         except (p.ParseException, IndexError):
-            if self.default_value is not self._optionalNotMatched:
+            if self.defaultValue is not self._optionalNotMatched:
                 if self.expr.resultsName:
-                    tokens = p.ParseResults([self.default_value])
-                    tokens[self.expr.resultsName] = self.default_value
+                    tokens = p.ParseResults([self.defaultValue])
+                    tokens[self.expr.resultsName] = self.defaultValue
                 else:
-                    tokens = [self.default_value]
+                    tokens = [self.defaultValue]
             else:
                 tokens = []
         return loc, tokens
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -1252,9 +1252,9 @@ class CompactSyntaxParser(object):
     units_ref = make_kw('units') - nc_ident
 
     # Setting the units for the independent variable
-    setTimeUnits = (make_kw('independent') - make_kw('var') - units_ref("units")).setParseAction(Actions.SetTimeUnits)
+    set_time_units = (make_kw('independent') - make_kw('var') - units_ref("units")).setParseAction(Actions.SetTimeUnits)
     # Input variables, with optional units and initial value
-    inputVariable = p.Group(
+    input_variable = p.Group(
         make_kw('input') -
         c_ident("name") +
         Optional(units_ref)("units") +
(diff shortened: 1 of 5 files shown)


MichaelClerx:
A few more bits and pieces. I think we can consider #19 closed after this one

MichaelClerx:
This one's not as scary as the previous one :D

fc/parsing/CompactSyntaxParser.py
```diff
@@ -204,9 +204,9 @@ def OperatorOperands(self):
                 except StopIteration:
                     break
 
-        def Operator(self, operator):
-            """Get the MathML element for the given operator."""
-            return getattr(M, self.OP_MAP[operator])
+        # def operator(self, operator):
```
jonc125: Can we just delete this, and the definition of `OP_MAP`?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1025,102 +1027,102 @@ class CompactSyntaxParser(object):
...
-    nullValue = p.Group(MakeKw('null')).setName('Null').setParseAction(Actions.Symbol('null'))
-    defaultValue = p.Group(MakeKw('default')).setName('Default').setParseAction(Actions.Symbol('defaultParameter'))
-    stringValue = quotedString.copy().setName('String').setParseAction(Actions.Symbol('string'))
+    null_value = p.Group(make_kw('null')).setName('Null').setParseAction(Actions.Symbol('null'))
+    default_value = p.Group(make_kw('default')).setName('Default').setParseAction(Actions.Symbol('defaultParameter'))
+    string_value = quoted_string.copy().setName('String').setParseAction(Actions.Symbol('string'))
 
     # Recognised MathML operators
-    mathmlOperators = set('''quotient rem max min root xor abs floor ceiling exp ln log
+    mathml_operators = set('''quotient rem max min root xor abs floor ceiling exp ln log
```
jonc125: Tidy up the indentation on the next few lines?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1025,102 +1027,102 @@ class CompactSyntaxParser(object):
...
                              sin cos tan   sec csc cot   sinh cosh tanh   sech csch coth
                              arcsin arccos arctan   arccosh arccot arccoth
                              arccsc arccsch arcsec   arcsech arcsinh arctanh'''.split())
 
     # Wrapping MathML operators into lambdas
-    mathmlOperator = (p.oneOf('^ * / + - not == != <= >= < > && ||') |
-                      p.Combine('MathML:' + p.oneOf(' '.join(mathmlOperators))))
-    wrap = p.Group(p.Suppress('@') - Adjacent(p.Word(p.nums)) + Adjacent(colon) + mathmlOperator
+    mathml_operator = (p.oneOf('^ * / + - not == != <= >= < > && ||') |
+                      p.Combine('MathML:' + p.oneOf(' '.join(mathml_operators))))
```
jonc125: Also here?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -1445,21 +1447,21 @@ def try_parse(self, callable, *args, **kwargs):
                                % (int(self._stack_depth_factor * self._original_stack_limit),))
         return r
 
-    # def ParseFile(self, filename, xmlGenerator=None):
+    # def parse_file(self, filename, xml_generator=None):
```
jonc125: We can probably delete this function?

test/test_CompactSyntaxParser.py
```diff
@@ -125,7 +125,7 @@ def test_parsing_simple_expressions(self):
                           [[['1', '<', '2'], ['3', '+', '4'], ['5', '*', '6']]])
 
     # XML-Only method?
-    # def testParsingTrace(self):
+    # def test_parsing_trace(self):
```
jonc125: Yeah, this is XML-only and can be removed. The Python implementation can trace naturally using the stack, so doesn't need special support!

test/test_CompactSyntaxParser.py
```diff
@@ -833,15 +833,15 @@ def test_parsing_post_processing(self):
 }
 """
         out = [[[['a'], [['check', ['sim:result']]]], [['a', '>', '5']]]]
-        self.assertParses(csp.postProcessing, mls, out)
+        self.assertParses(csp.post_processing, mls, out)
 
     '''
     def test_parsing_full_protocols(self):
```
jonc125: I reckon we can delete this test too.

fc/parsing/CompactSyntaxParser.py
MichaelClerx: ok!

fc/parsing/CompactSyntaxParser.py
MichaelClerx: will try!

fc/parsing/CompactSyntaxParser.py
MichaelClerx: will try1

fc/parsing/CompactSyntaxParser.py
MichaelClerx: ok!

test/test_CompactSyntaxParser.py
MichaelClerx: Do we still need the `class Trace` and other tracing methods then?

test/test_CompactSyntaxParser.py
MichaelClerx: done!

test/test_CompactSyntaxParser.py
jonc125: Ah, sorry, I misremembered. There is still something that happens in the Python implementation with this - it sets `wrapped_expr.trace = True`, which triggers a call to `Locatable.trace_value` when the expression is evaluated. It's a debugging aid, though probably isn't used much!

You could enhance the test with a check like this?
```python
action = grammar.parseString('var?', parseAll=True)
assert action.expr().trace

action = grammar.parseString('var', parseAll=True)
assert not action.expr().trace
```

MichaelClerx: So what would I do with the PROTO_NS bit?

```
    def testParsingTrace(self):
       trace = {'{%s}trace' % CSP.PROTO_NS: '1'}
       self.assertParses(csp.expr, '1?', [['1']])
       self.assertParses(csp.expr, 'var?', [['var']])
       self.assertParses(csp.expr, '(1 + a)?', [[['1', '+', 'a']]])
       self.assertParses(csp.expr, '1 + a?', [['1', '+', ['a']]])
```

jonc125: That bit can go - it was just checking that the XML conversion added the attribute.

MichaelClerx:
```
        action = csp.expr.parseString('var?', parseAll=True)
        print(action)
        print(type(action))
        assert action._expr().trace
```
I've tried this but that returns:
```
[Trace[Variable[var]]]
<class 'pyparsing.ParseResults'>
```
and then
```
    assert action._expr().trace
E   TypeError: 'str' object is not callable
```

jonc125: Maybe you need `action[0].expr().trace` then? Looks like the `Action` class is wrapped from the outer `[]`

Issue #49, opened 2019-09-11
Issue with pyparsing 2.4.2


MichaelClerx:
Causes infinite recursion when importing compact syntax parser.

MichaelClerx:
https://github.com/pyparsing/pyparsing/issues/123

PR #60, opened 2019-09-25
Further monkey patching to pyparsing to avoid 2.4.x infinite loop.

MichaelClerx:
Possible fix to #49

### DIFF
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -984,6 +984,9 @@ def MonkeyPatch():
 
     def ignore(self, other):
         """Improved ignore that avoids ignoring self by accident."""
+        if isinstance(other, str):
+            other = Suppress(other)
+
         if isinstance(other, p.Suppress):
             if other not in self.ignoreExprs and other != self:
                 self.ignoreExprs.append(other.copy())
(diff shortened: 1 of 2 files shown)


MichaelClerx:
Not sure if this would break previous versions. Maybe better to keep disallowing 2.4.x branch, and wait for the issue to be fixed upstream?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -984,6 +984,9 @@ def MonkeyPatch():
 
     def ignore(self, other):
         """Improved ignore that avoids ignoring self by accident."""
+        if isinstance(other, str):
+            other = Suppress(other)
```
jonc125: Should this be `p.Suppress`?

fc/parsing/CompactSyntaxParser.py
```diff
@@ -984,6 +984,9 @@ def MonkeyPatch():
...
+
         if isinstance(other, p.Suppress):
             if other not in self.ignoreExprs and other != self:
```
jonc125: Maybe making the second test here `other is not self` would suffice for all versions? Actually, no, I suspect we do need the equality check because you could potentially have multiple `Suppress` objects wrapping the same underlying expression.

jonc125:
Looks like the underlying issue will be fixed in 2.4.3 so probably best not to merge this but wait until the release and then update our pin?

MichaelClerx:
https://github.com/pyparsing/pyparsing/issues/123

Do I get some kind of badge for contributing to an open source project I'm not involved in :D :D :D

jonc125:
You're a few days early for it to contribute to Hacktoberfest :)

PR #66, opened 2019-10-29
Missing units in unit registry fixes #65

### LINKED ISSUE #65: Add protocol units to cellmlmanip unit registry
Using a model with units seconds (earm_noble_model_1990) and a protocol with units milliseconds causes a fail in unit conversion as the unit_registry does not know about ms. 

Initially incorrectly filed in https://github.com/ModellingWebLab/cellmlmanip/issues/110

This arises from https://github.com/ModellingWebLab/weblab-fc/issues/22

skeating:


skeating:
Deleting this pull request as it was my mistake that made me think it was needed. Few typographical changes have been merged in 22-graphstate branch

Issue #64, opened 2019-10-24
Run tests on Python3.5


MichaelClerx:
To stick with cg and cellmlmanip

MichaelClerx:
Pytables won't run on 3.5, see #67

MichaelClerx:
Requiring 3.6 from now on! (Discussion in #67)

PR #67, opened 2019-11-05
Adding python 3.5 testing

MichaelClerx:
See #64

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python
 cache: pip
 python:
-  - "3.6"
+  - "3.5"
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


MichaelClerx:
This fails due to an hdf5 issue. Seems pytables doesn't test on 3.5: https://github.com/PyTables/PyTables/blob/master/appveyor.yml

I'd suggest dropping the 3.5 requirement for weblab-fc then?

jonc125:
Interesting! I'm sure I've used pytables on 3.5 before, but probably installed in a conda env which will get around the building from source error travis is seeing. Looks like there's no wheel available for that combination.

I'm happy to target 3.6+ for weblab-fc though.

PR #68, opened 2019-11-05
Update pyparsing version requirements

MichaelClerx:
See #49 

New version 2.4.3 is out, fixes issue. It worked with older versions too, so now setup.py just disallows 2.4.2


MichaelClerx:
Built this on the 3.5 testing branch, creating a new PR!

PR #69, opened 2019-11-05
Updated version requirement for pyparsing.

MichaelClerx:
See #49

New version 2.4.3 is out, fixes issue. It worked with older versions too, so now setup.py just disallows 2.4.2


MichaelClerx:
Tests fail due to change in cellmlmanip, but the pyparsing bit works!

Issue #33, opened 2019-04-18
Move Load2d out of test_support?


MichaelClerx:
`fc.language.values.LoadFunction` depends on this method, but is not a test.

It would be nice to completely separate these things, and then move test_support into the test module

MichaelClerx:
Closed in #70

PR #70, opened 2019-11-05
Moved data loading methods into own module, outside of test_support.

MichaelClerx:
See #33

### DIFF
--- a/fc/language/values.py
+++ b/fc/language/values.py
@@ -1,6 +1,12 @@
+"""
+Classes for values in the protocol language (e.g. numbers, but also lambda
+functions and data loading).
+"""
 
+import os
 import numpy as np
 
+from ..data_loading import load2d
 from ..error_handling import ProtocolError
 
 
@@ -172,8 +178,6 @@ class LoadFunction(LambdaClosure):
             raise ProtocolError("A load() call takes a single parameter, not %d." % len(actual_parameters))
         if not isinstance(actual_parameters[0], String):
             raise ProtocolError("A load() call takes a string parameter with the file path to load.")
-        import os
         file_path = os.path.join(self.base_path, actual_parameters[0].value)
-        from ..test_support import load2d
         return load2d(file_path)
 
--- a/fc/language/values.py
+++ b/fc/language/values.py
@@ -6,8 +6,8 @@ functions and data loading).
 import os
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


MichaelClerx:
The failing test is unrelated (and also fixed now, so re-running would clear this up but I lack the permissions :D )

fc/language/values.py
```diff
@@ -1,6 +1,12 @@
+"""
+Classes for values in the protocol language (e.g. numbers, but also lambda
+functions and data loading).
+"""
 
+import os
 import numpy as np
 
+from ..data_loading import load2d
```
jonc125: This could cause issues as it sets up a circular import.

MichaelClerx: What's your preferred solution? I stick the import in the method that uses it, or change it to `from .. import data_loading`
I feel the second is slightly cleaner?

jonc125: The former is the canonical solution to such problems I believe (and what was done before). Is the latter a new solution in Python 3 (I don't think it worked in Python 2 at least)?

MichaelClerx: Hmmmm! I thought the 2nd worked because the sequence of events would be

- start importing module X
- X imports Y, so start importing Y
- Y imports X, so Y gets a pointer to the unfinished module X
- Some method in Y uses currently unknown bits of X, but Python doesn't care yet as it doesn't check the function contents until you call it
- Y finishes importing
- X gets a pointer to Y, and continues importing

So it would be ok because nothing at the top level in Y accessed members of X, just X itself
(which means there's a difference between import X and from X import Z)

MichaelClerx: I _think_ that's how it works. Don't quote me on that!

MichaelClerx: Some discussion here https://stackoverflow.com/questions/744373/circular-or-cyclic-imports-in-python

jonc125: Yeah, that's basically how it works, so you're safe if your modules basically just define classes & functions (and import the whole module rather than individual bits from it), but if they "do" anything during import you can get errors. I think this used to bite with the syntax parser especially because that did more stuff at import time; with the removal of the XML support we might not hit this as easily!

So actually given tests pass we could keep things as they are for now. Or do `from .. import data_loading` if you prefer to be safer.

Issue #32, opened 2019-04-18
Document use of setproctitle, or drop it


MichaelClerx:
Some tests include:

```
try:
    from setproctitle import setproctitle
except ImportError:
    def setproctitle(t):
        return None
```

This optional dependency should be documented somewhere

PR #71, opened 2019-11-05
Made setproctitle a test dependency (not a dev one).

MichaelClerx:
See #32. It's only used in a test

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -90,12 +90,12 @@ setup(
     extras_require={
         'dev': [
             # 'line_profiler',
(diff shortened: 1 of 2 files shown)


jonc125:
But it's only *useful* when running the tests locally, as it lets you see which process is running which model/protocol combination, e.g. using top, iirc.

MichaelClerx:
But you install things for local testing with the test package?

MichaelClerx:
But ok, I see your point. Have reversed & added a note

Issue #30, opened 2019-03-22
Run `test_while_loop` with a generated model


MichaelClerx:
See #23 

Test_while_loop adds an input and overrides an existing equation

- [x] Parse model inputs --> Get info down to model generation stage in FC
  - [x] Update code generation with inputs in CG
- [x] Parse equation --> Store equation somehow in ~CG~ FC
  - [x] Convert to sympy and resolve variable references (requires cellmlmanip model object)
  - [x] Raise (friendly) error if references can't be resolved
  - [x] Check if equation uses FC parts that don't fit in sympy?? (e.g. arrays)
  - [x] Update model with new equation

MichaelClerx:
@jonc125 is that the logic you'd expect above?
Not sure how to get the equation info at the moment. Would mean either having some intermediate (unresolved) equation coming back from CompactSyntaxParser or passing the model (or a dict of name:dummy pairs) into the parser

jonc125:
Some points:
* You shouldn't interact with CG until all the manipulation (add input, change equation) is done in cellmlmanip. So modify the model, *then* generate code.
* Equations in the model interface section can't use FC-specific constructs, only things allowed in CellML (except for the magic 'interpolate on a data file' construct, which is in a later test case).
* Not sure on the easiest approach to parsing. Would Sympy be able to parse the raw equation string? The challenge there would be keeping the units on numbers (and possibly recognising prefix:term style variable names). Or use the existing parse tree from pyparsing and have a mapping from that to Sympy terms?
* Name references would need to be resolved at this parsing stage so you get the correct VariableDummy. Which is why you handle input/output/var declarations before equations, so all symbols you need have been created. A dict of name:dummy pairs isn't quite what you need, because (most) of the name references in the protocol are to annotations (except for purely local-to-protocol declared variables).

MichaelClerx:
> You shouldn't interact with CG until all the manipulation (add input, change equation) is done in cellmlmanip. So modify the model, then generate code.

Woops, meant to write FC there!

> Equations in the model interface section can't use FC-specific constructs, only things allowed in CellML (except for the magic 'interpolate on a data file' construct, which is in a later test case).

Is this handled by the CompactSyntaxParser?

MichaelClerx:
Is it what this is for?
https://github.com/ModellingWebLab/weblab-fc/blob/master/fc/parsing/CompactSyntaxParser.py#L1143

MichaelClerx:
[...]

Issue #22, opened 2019-02-18
Run GraphState on models requiring units conversion


MichaelClerx:
Continue with the GraphState protocol, but use models that require units conversion on the outputs. Since GraphState only has outputs (no inputs) a conversion for V is the simplest possible case. Converting time is a little more complicated.
- [x] The paci models have V in volts (and are indeed the only ones not to use mV!)
- [x] aslanidi_atrial_model_2009 has time in secs, as do noble models except 1962, and some others. Pick a simple one :)
- [x] ~~A good test case for all the units conversion stuff will be the test_simple_odes model in cellmlmanip, actually. Though we don't have reference data for this at present.~~

Doing the V case turns out to require handling the state_variable annotation properly, as otherwise your state vector output contains the unconverted version of V!
- [x] Before units converting outputs, annotate all state variables (use `Model.get_state_symbols` etc)
- [x] Implement the equivalent of `process_output_variable_vectors`
- [x] Handle vector outputs in code generation
- [x] We may need a `Model.remove_variable` method to get rid of unused variables and associated RDF annotations, so unused state variables don't get pulled in by accident!

PR #73, opened 2019-11-07
Parse more parts of the model interface

MichaelClerx:
For
- Running `graphstate` on models requiring unit conversion #22 
- Running `test_while_loop` on a generated model #30


MichaelClerx:
https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/fc/protocol.py#L624-L666

@jonc125 do the comments I put in here make sense?

@skeating this is where we do the model manipulation - and I think input/output unit conversion - in `fc`. One difficulty is that the information to do this (the ModelInterface class) didn't exist before (pycml read the protocol too) so I'm extracting required info from the parser and sticking it in the ModelInterface as I go along

jonc125:
> do the comments I put in here make sense?

They're along the right lines. Some thoughts quickly:

* "Create protocol unit store" is probably best done when the protocol is parsed/created - it's not model-specific, and indeed should be independent of the model
* Outputs can be of any type. And handling their units will be done by cellmlmanip's `add_output` method.
* I don't think the state_variable output should be handled specially here; just treat it as a standard ontology term (but like concentration, it can annotate multiple variables: vector outputs in pycml terms).
* The `define` statement handling should also modify the model, but I guess that's implicit?
* And of course there are some steps missing, but we're not worrying about those yet.

MichaelClerx:
Some notes after a chat with @skeating :

- This branch/PR can be repurposed as the unit-conversion branch, as it contains some work on extracting the required info from the model interface and getting it to the cellmlmanip model
- There's a test in the branch called [test_generated_model_while_loop](https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/test/test_generated_model_while_loop.py) that doesn't terminate, so will have to be disabled/removed for unit conversion work

- Information from the FC protocol is parsed in the [CompactSyntaxParser](https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/fc/parsing/CompactSyntaxParser.py), but not all required info is currently read in (because pycml used to do this).
- Instead, the required info can be seen in the commented out `_xml` methods in the CompactSyntaxParser, and transferred to the `_expr` methods
- The _expr method returns unstructured dicts, see e.g. https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/fc/parsing/CompactSyntaxParser.py#L482-L509

- In `protocol.py`, the info from the parser is combined (because protocols can import other protocols, so there's several parsing trees potentially), and then stored in the protocol
- I've started making a more structured class for the [ModelInterface data](https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/fc/protocol.py#L750). 
  - This may or may not be a good idea, maybe a style thing? Up to you to expand, continue, or ditch this approach!

- The cellmlmanipmodel is created in [protocol.set_model](https://github.com/ModellingWebLab/weblab-fc/blob/30-test-while-loop/fc/protocol.py#L592)
- I've outlined a basic scheme of how to read inputs and where to do unit conversion here (see also discussion above)

Hope that's useful! @jonc125 feel free to directly modify this message if there's any major mistakes :-)

jonc125:
I'm not going to push any changes to this branch as I'm not sure whether @skeating has any uncommitted code. But some thoughts:
* The `ModelInterface` class etc. should probably move to a new file in the `parsing` subfolder? And then the parser _expr methods can return instances of these directly?
* Seems like `exposeNamedParameters` is no longer used so we can get rid of that!

I could work on these, along with starting to parse the `define` stuff, in a separate branch building off this?

MichaelClerx:
Sounds great! Or you could continue in this branch I guess (Sarah mentioned she didn't get very far)

jonc125:
@MichaelClerx I forgot to create a separate branch before pushing https://github.com/ModellingWebLab/weblab-fc/pull/73/files/bc4cf9af6050a66bd25f676d010303253b3446a9..716bdaa4842a25150c39dc1ef2e38aa6edc9e535, but they're actually working on the topic of this PR, so ... :)

Does this look like a reasonable approach to be taking?

MichaelClerx:
Yup, that looks sensible to me!

jonc125:
@MichaelClerx I think we could merge this, except some units stuff [is failing](https://travis-ci.com/ModellingWebLab/weblab-fc/builds/147761160) :( Looks like I'm not allowed to pass e.g. 'millivolt' as a unit definition string, which is surprising me. Any suggestions?

jonc125:
Hooray! Graph state and while loop protocols pass (albeit without unit conversions)!

PR #80, opened 2019-11-14
Add some notes on how pycml manipulates models

jonc125:


pycml/README.md
```diff
@@ -2,3 +2,44 @@
...
+- arising from this, managing name references in mathematics (which can be local names, fully qualified names, or ontology terms) particularly when equations are moved or changed;
+- the need for units to be defined in a component and referenced by name from variables;
+- the internal state needed by code generation, for instance to classify variables (computed, state, etc) and topologically sort them. This state can shift as model modifications are made, and might not be fully known until all modifications have been done, so the state can sometimes be 'nonsense' at intermediate stages.
+
+Protocol parsing starts with `parse_protocol`, called by `apply_protocol_file`. It ensures that the different parts of the interface are processed in an order that respects the principle of least surprise for protocol authors: it tries to ensure their modifications do what they probably expect!
+
+Imports are treated first, included those arising implicitly from nested protocols. Where relevant, units definitions, interface definitions, etc. from imports are merged with the parent protocol's definitions.
```
MichaelClerx: Does "arising implicitly" here mean "explicitly imported but in an imported protocol", or is there some other way that they can arise?

jonc125: Gory details are in the method, but in short, I think it's just model modifications (i.e. the model interface section) that matter from nested protocols.

PR #83, opened 2020-01-22
FileNotFound should return a clean exception

### LINKED ISSUE #82: Attempting to create a protocol from non existent file
There are possibly two issues here:

1. If you call a Protocol with an non-existent file you do not get a nice exception. You get both FileNotFoundError and AttributeError from pyparsing as it throws one while dealing with other.
 So this test does not pass:
   def test_non_existent(self):
        proto_file = 'protocols/not_here.txt'
        self.assertRaises(FileNotFoundError, fc.Protocol(proto_file))

Are  we worried ??

2.  If the protocol tries to import from the library files and for some reason doesnt find them then it just cleanly exits with no error at all. This one did worry me. Should we address this.

skeating:
## Description
Some unfound files throw several errors and some just exit quietly

## Motivation and Context

Fixes #82

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I have updated all documentation necessary.
- [x] I have checked spelling in (new) comments.

## Testing
- [X] Testing is done automatically and codecov shows test coverage
- [ ] This cannot be tested automatically 

## TO DO
- [x] protocol file not found
- [x] imported library file not found

### DIFF
--- a/fc/protocol.py
+++ b/fc/protocol.py
@@ -81,6 +81,19 @@ class Protocol(object):
 
         The protocol must be specified using the textual syntax, as defined by the CompactSyntaxParser module.
         """
+        # if the filename passed as argument to CompactSyntaxParser is not found
+        # it tries to read it before checking it exists
+        # and you get a messy exit
+        # file_contents = file_or_filename.read()
+        #     E AttributeError: 'str' object has no attribute 'read'
+        #
+        #     During handling of the above exception, another exception occurred:
+        #     E FileNotFoundError: [Errno 2] No such file or directory:
+        # throw exception
+
+        with open(proto_file) as f:
+            f.close()
+
         self.proto_file = proto_file
         self.proto_name = os.path.basename(self.proto_file)
 
--- a/test/protocols/test_non_existent_library.txt
+++ b/test/protocols/test_non_existent_library.txt
@@ -1,0 +1,39 @@
+# Test using the original definition when replacing an equation.
+
+namespace oxmeta = "https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#"
+namespace test = "urn:test-ns#"
+
+import std = "BasicLibrary.txt"
+
+units {
+    mV = milli volt
+    ms = milli second
+    mV_per_ms = mV . ms^-1
+}
+
+model interface {
+    output oxmeta:membrane_voltage units mV
+    output test:parameter_a
+    output test:parameter_b
+
+    define diff(oxmeta:membrane_voltage; oxmeta:time) = 10 :: mV_per_ms + original_definition  # i.e. 10 + a
+    define test:parameter_b = (5 :: dimensionless * original_definition) - original_definition # i.e. (5*(-1)) - (-1) = -4
+    define test:parameter_a = 1 :: mV_per_ms + original_definition
+}
+
+tasks {
+    simulation sim = timecourse {
+        range t units ms uniform 0:5
+    }
+}
+
+post-processing {
+    assert std:Close(sim:membrane_voltage, [12*i for i in 0:6])
+    assert sim:parameter_a[0] == 2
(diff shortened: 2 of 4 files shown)


skeating:
I assume there is no automatic testing since I'm only trying to merge into another branch

jonc125:
That's odd! The first commit got tested on Travis, and I'd have thought all of them should be. Certainly it tests PRs that merge to a branch in the WebLab repo...

fc/protocol.py
```diff
@@ -81,6 +81,19 @@ def __init__(self, proto_file, indent_level=0):
...
+        # it tries to read it before checking it exists
+        # and you get a messy exit
+        # file_contents = file_or_filename.read()
+        #     E AttributeError: 'str' object has no attribute 'read'
+        #
+        #     During handling of the above exception, another exception occurred:
+        #     E FileNotFoundError: [Errno 2] No such file or directory:
+        # throw exception
+
+        with open(proto_file) as f:
```
jonc125: Would it be better to do something like `assert os.path.exists(proto_file)`? That doesn't check you can read it, but seems cleaner than just trying to open the file! Or put more checks where the file is read?

test/test_syntax_interface.py
```diff
@@ -664,6 +664,24 @@ def test_graph_txt(self):
         proto.set_model(Model.TestOdeModel(1))
         proto.run()
 
+    def test_non_existent(self):
+        proto_file = 'protocols/not_here.txt'
+        # for some reason this line reports failure (while also reporting the Exception !!
+        # self.assertRaises(FileNotFoundError, fc.Protocol(proto_file))
```
jonc125: If you're using the unittest assertRaises, you need to call it like `self.assertRaises(FileNotFoundError, fc.Protocol, proto_file)`. Or since we're actually running via pytest, we can use `with pytest.raises(FileNotFoundError): ...`

fc/protocol.py
skeating: I was specifically going for getting the FileNotFound exception rather a generic one but will defer to whichever you wish :-)

jonc125: Fair enough. And actually I'm not sure there's a simpler way to check for readability of a file!

test/test_syntax_interface.py
skeating: okay BUT the rest of the file uses the unittests syntax

jonc125: Yeah, that's because it was written before pytest existed :) I'd like to migrate over gradually, so at least new stuff it written more in the pytest style.

fc/protocol.py
skeating: Putting the checks where the file is read is difficult as the CompactSyntaxParser(CSP) code uses pyparsing to do this and in a fairly inconsistent way where is tries to open before checking it exists and gets an exception that throws an exception. 
I was trying to avoid needing to look closer at CSP.
Also when looking for a library file I need to do this catch before it tries to do anything with the protocol. Locating the check at the beginning of Protocol init catches both cases neatly; trying to hack CSP does not.

test/test_parsing_errors.py
```diff
@@ -0,0 +1,21 @@
+
+import pytest
+
+import fc
+
+
+def test_non_existent():
+    proto_file = 'protocols/not_here.txt'
```
MichaelClerx: One-liner comments saying what this test is for?

test/protocols/test_non_existent_library.txt
```diff
@@ -0,0 +1,39 @@
+# Test using the original definition when replacing an equation.
```
MichaelClerx: Comment copied from previous test file?

PR #84, opened 2020-02-03
Some syntax parser tidying

### LINKED ISSUE #26: Split CompactSyntaxParser into multiple files
For example, actions can easily go into one or multiple files + get some more documentation

Not sure we still need to be able to run it as a script either...

jonc125:
I decided to fix #26 while getting back up to speed on the parser code...


fc/protocol.py
```diff
@@ -205,16 +205,16 @@ def _parse(self):
...
-        CSP.Actions.source_file = self.proto_file
         generator = self.parsed_protocol = parser.try_parse(
             CSP.CompactSyntaxParser.protocol.parseFile,
             self.proto_file,
             parseAll=True
         )[0]
-        assert isinstance(generator, CSP.Actions.Protocol)
-
-        # A class:`fc.parsing.CompactSyntaxParser.Actions.Protocol` object,
+        # A class:`fc.parsing.actions.Protocol` object,
```
MichaelClerx: Missing a `:` in front of "class", but doesn't really need RST as it's not a docstring anyway

Issue #85, opened 2020-02-03
Add templating code and WebLabPrinter


MichaelClerx:
See https://github.com/ModellingWebLab/chaste-codegen/issues/54

PR #86, opened 2020-02-03
Adding code generation code and templates

MichaelClerx:
See #85

- Removed dependency on `weblab_cg`
- Copied in templates (including temporary ones for reference)
- Copied in weblab printer, and template processing code
- Copied in tests
- Updated HH model with more annotations (used in tests for code generation)
- Added local version of `_exp` hack for sympy


fc/templates/unused/TestClampingToInitialValue.pyx
```diff
@@ -0,0 +1,135 @@
+# cython: profile=True
```
jonc125: I suspect these were useful when you were first figuring out the template, but aren't really needed now. However, let's keep them in this PR and delete when we're a bit further forward with all the code generation features.

fc/code_generation.py
```diff
@@ -0,0 +1,284 @@
...
+            'var_name': var_name,
+            'parameter_index': parameter_index,
+            'length': length,
+        })
+
+    # Create RHS equation information dicts
+    rhs_equations = []
+    for eq in model.get_equations_for(model.get_derivative_symbols()):
+        # TODO: Parameters should never appear as the left-hand side of an
+        # equation (cellmlmanip should already have filtered these out).
```
jonc125: This is probably not true now... (but probably shouldn't fix in this PR?)

fc/code_generation.py
```diff
@@ -0,0 +1,284 @@
...
+    for i, parameter in enumerate(parameters):
+        symbol = model.get_symbol_by_ontology_term(*parameter)
+        parameter_info.append({
+            'index': i,
+            'annotation': parameter,
+            'var_name': symbol_name(symbol),
+            'initial_value': model.get_value(symbol),
+        })
+
+    # Create map of parameter symbols to their indices
```
jonc125: Merge this loop with the one above?

fc/templates/weblab_model.pyx
```diff
@@ -0,0 +1,347 @@
...
+    #        Sundials.N_VDestroy_Serial(self._parameters)
+
+    def get_environment_map(self):
+        """
+        Get a map from ontology prefix to the environment containing model
+        variables annotated with that ontology.
+
+        See :meth:`fc.simulations.AbstractOdeModel.get_environment_map()`.
+        """
+        # TODO Some part of this might need to be generated
```
jonc125: Yeah, eventually it all should be, based on ontologies actually used in the protocol.

fc/code_generation.py
MichaelClerx: Thanks. Made an issue for it https://github.com/ModellingWebLab/weblab-fc/issues/87

fc/templates/unused/TestClampingToInitialValue.pyx
MichaelClerx: Agreed! https://github.com/ModellingWebLab/weblab-fc/issues/88

PR #89, opened 2020-02-05
Migrate ModelInterface functionality to parse actions

jonc125:
This PR contains enough to get the GraphState protocol working again without units conversions.

It also updates a few things to the latest cellmlmanip.

### DIFF
--- a/fc/code_generation.py
+++ b/fc/code_generation.py
@@ -218,7 +218,7 @@ def create_weblab_model(path, class_name, model, outputs, parameters):
 
         # Allow special output value 'state_variable'
         # TODO Replace this with a more generic implementation
-        if output == 'state_variable':
+        if output[1] == 'state_variable':
             var_name = [{'index': x['index'], 'var_name': x['var_name']}
                         for x in state_info]
             parameter_index = None
@@ -251,7 +251,7 @@ def create_weblab_model(path, class_name, model, outputs, parameters):
     # Create output equation information dicts
     output_equations = []
     output_symbols = [model.get_symbol_by_ontology_term(*x) for x in outputs
-                      if x != 'state_variable']
+                      if x[1] != 'state_variable']
     for eq in model.get_equations_for(output_symbols):
         output_equations.append({
             'lhs': printer.doprint(eq.lhs),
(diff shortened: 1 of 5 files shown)


fc/code_generation.py
```diff
@@ -220,7 +218,7 @@ def derivative_name(deriv):
 
         # Allow special output value 'state_variable'
         # TODO Replace this with a more generic implementation
-        if output == 'state_variable':
+        if output[1] == 'state_variable':
```
MichaelClerx: I didn't realise `state_variable` had a URI?

fc/parsing/actions.py
```diff
@@ -483,46 +488,55 @@ def _expr(self):
...
-    # def _xml(self):
-    #     return P.setIndependentVariableUnits(**self.get_attribute_dict('units'))
-    pass
+    """Parse action for specifying the units time should have in the model.
+
+    ``independent var units <uname>``
+    """
+    def _expr(self):
+        self.time_units = self.get_named_token_as_string('units')
+        assert self.time_units is not None
```
MichaelClerx: Is this a sanity check, or can this show up in the user's error output?

fc/parsing/actions.py
```diff
@@ -610,20 +636,58 @@ class UnitsConversion(BaseGroupAction):
...
-            InputVariable,
-            ModelEquation,
-        )
-        for action in self:
-            if isinstance(action, handled):
-                output.append(action.expr())
-            # TODO: Create objects for all parts of the model interface
-        return output
+        actions = self.get_children_expr()
+        print('in MI._expr', self.tokens, actions, map(type, actions))
```
MichaelClerx: print!

fc/code_generation.py
jonc125: Yep, it's a standard oxmeta:state_variable reference.

fc/parsing/actions.py
jonc125: Should be just a sanity check - the parser would complain before now if it didn't exist.

fc/parsing/actions.py
jonc125: Oops!

PR #90, opened 2020-02-06
Parse units definitions from protocol

jonc125:


jonc125:
This won't pass tests until https://github.com/ModellingWebLab/cellmlmanip/pull/215 is merged, but illustrates how we can use the new units API @MichaelClerx - looks quite nice to me!

fc/protocol.py
```diff
@@ -151,57 +152,58 @@ def __init__(self, proto_file, indent_level=0):
 
         # 4. Any number of ``import`` statements (again, no section)
         # https://chaste.cs.ox.ac.uk/trac/wiki/FunctionalCuration/ProtocolSyntax#Importsofotherprotocols
-
-        # TODO: Not sure what this dict contains
+        # Maps an import 'name' prefix to a :class:`Protocol` instance.
```
MichaelClerx: Thanks for all these!

fc/protocol.py
```diff
@@ -151,57 +152,58 @@ def __init__(self, proto_file, indent_level=0):
...
-
-        # TODO: Do all of these end up in this list?
-        # TODO: Who checks the assertions?
+        # These statements will be executed when the protocol is run.
         self.library = []
 
         # 6. The ``units`` section
         # https://chaste.cs.ox.ac.uk/trac/wiki/FunctionalCuration/ProtocolSyntax#Physicalunitdefinitions
-        # TODO: Where do these end up?
+        self.units = UnitStore()
```
MichaelClerx: would have to be `self.units = UnitStore(model.units)`, I guess?

fc/parsing/actions.py
```diff
@@ -941,60 +941,77 @@ def _expr(self):
...
+            raise ValueError('Offset units are no longer supported')
+        self.units = self.get_named_token_as_string('units')
+        assert self.units is not None
+        self.prefix = self.get_named_token_as_string('prefix', '')
+        self.exponent = self.get_named_token_as_string('exponent')
+        self.multiplier = self.get_named_token_as_string('multiplier')
+
+        expr = self.prefix + self.units
+        if self.exponent is not None:
+            expr = '{}**{}'.format(expr, self.exponent)
```
MichaelClerx: Do any of these need brackets?

fc/protocol.py
jonc125: Except we're setting up the protocol first, so actually when parsing the model we want to tell it to use the protocol's store?

fc/parsing/actions.py
jonc125: I don't /think/ so, precedence should make it work...

fc/protocol.py
MichaelClerx: Guess so!

Issue #77, opened 2019-11-07
Move plotting code out of protocol class


MichaelClerx:
Seems like it has enough responsibilities without also doing this.
Plus would get rid of the distracting matplotlib imports

jonc125:
Fine by me.

PR #92, opened 2020-02-07
Move plotting code out of Protocol class.

MichaelClerx:
See #77.

### DIFF
--- a/fc/plotting.py
+++ b/fc/plotting.py
@@ -4,7 +4,7 @@ Methods for plotting simulation results.
 import operator
 from functools import reduce
 
-from .error_handling import ProtocolError  # noqa: E402
+from .error_handling import ProtocolError
 
 import matplotlib
 matplotlib.use('agg')
--- a/fc/plotting.py
+++ b/fc/plotting.py
@@ -1,0 +1,53 @@
+"""
+Methods for plotting simulation results.
+"""
+import operator
+from functools import reduce
+
+from .error_handling import ProtocolError  # noqa: E402
+
+import matplotlib
+matplotlib.use('agg')
+import matplotlib.pyplot as plt  # noqa: E402
+plt.switch_backend('agg')  # on some machines this is required to avoid "Invalid DISPLAY variable" errors
+
+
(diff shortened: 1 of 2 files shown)


fc/plotting.py
```diff
@@ -0,0 +1,53 @@
+"""
+Methods for plotting simulation results.
+"""
+import operator
+from functools import reduce
+
+from .error_handling import ProtocolError  # noqa: E402
```
jonc125: Probably don't need the noqa for this one?

fc/protocol.py
```diff
@@ -27,6 +20,7 @@
 from .language.statements import Assign  # noqa: E402
 from .locatable import Locatable  # noqa: E402
 from .parsing import actions  # noqa: E402
+from .plotting import create_plot  # noqa: E402
```
jonc125: And now we've moved the matplotlib import, all these noqa markers should be redundant.

fc/protocol.py
```diff
@@ -445,53 +439,30 @@ def outputs_and_plots(self, errors, verbose=True, write_out=True):
         # Plots
         if write_out:  # suppress plotting when performing fitting
             start = time.time()
-            for plot in self.plots:
-                with errors:
+            with errors:
```
jonc125: I think doing this the other way around means you try each plot, even if earlier ones had errors. Which is what we want.

fc/protocol.py
```diff
@@ -445,53 +439,30 @@ def outputs_and_plots(self, errors, verbose=True, write_out=True):
...
                             'plotting', plot['title'],
                             'curve:', plot_descriptions[plot['y']],
                             'against', plot_descriptions[plot['x']]
                         )
-                    x_data = []
-                    y_data = []
-                    x_data.append(self.output_env.look_up(plot['x']).array)
-                    y_data.append(self.output_env.look_up(plot['y']).array)
+
+                    # TODO Looks like this isn't used?
```
jonc125: Indeed, it doesn't go in the .png, but having it here means we check for the error case.

PR #93, opened 2020-02-10
GraphState protocol on paci model

### LINKED ISSUE #22: Run GraphState on models requiring units conversion

### LINKED ISSUE #30: Run `test_while_loop` with a generated model

jonc125:
This tests the voltage unit conversion part of #22. Will close #22 and close #30.

The model manipulation code structure is now much nicer too.


jonc125:
This will benefit from refactoring into helper methods etc before merging I think @MichaelClerx , but is probably ok for an initial look. (There's also some debug output still in there!)

fc/code_generation.py
```diff
@@ -131,7 +135,39 @@ def uname(name):
     return symbols
 
 
-def create_weblab_model(path, class_name, model, outputs, parameters):
+_ONTOLOGY = None
+
+
+def get_variables_transitively(model, term):
```
MichaelClerx: This could do with a docstring!
It's not obvious to me what this method does from the names (what transitive about it?)

fc/code_generation.py
```diff
@@ -131,7 +135,39 @@ def uname(name):
...
+        cmeta_ids.update(model.rdf.subjects(pred_is_ver, annotation))
+    symbols = []
+    for cmeta_id in cmeta_ids:
+        assert isinstance(cmeta_id, rdflib.URIRef), 'Non-resource {} annotated.'.format(cmeta_id)
+        cmeta_id = str(cmeta_id)
+        if cmeta_id[0] != '#':
+            # TODO This should eventually be implemented
+            raise NotImplementedError(
+                'Non-local annotations are not supported.')
+        symbols.append(model.get_symbol_by_cmeta_id(cmeta_id[1:]))
```
MichaelClerx: Should we not be storing the info about how we found these variables somewhere? Or is it cheap enough to do that in a separate pass?

fc/protocol.py
```diff
@@ -670,10 +671,32 @@ def set_model(self, model):
                 model.add_equation(eq)
             # TODO: Check units of newly added equations; apply conversions where needed?
 
+            # Annotate state variables with the magic oxmeta:state_variable term
+            from cellmlmanip.rdf import create_rdf_node
+            is_version_of = create_rdf_node(('http://biomodels.net/biology-qualifiers/', 'isVersionOf'))
```
MichaelClerx: "version" seems a weird way to annotate this? "x[i]" is a version of "x"? Or "V" is a version of a state variable?

test/test_generated_model_graphstate.py
```diff
@@ -46,7 +46,34 @@ def test_graphstate_time_conversion():
...
+        'test', 'protocols', 'generated_model_graphstate.txt'))
+
+    # Set model (generates & compiles model)
+    model_name = 'paci_hyttinen_aaltosetala_severi_ventricularVersion'  # has voltage in volt, not millivolt
+    proto.set_model(os.path.join('test', 'models', model_name + '.cellml'))
+
+    # Run protocol
+    proto.set_output_folder('test_graphstate_voltage_conversion')
     proto.run()
     # Some test assertions are within the protocol itself
```
MichaelClerx: Do we need to do something here to check the output was sensible? (i.e. that the conversion actually took place)

MichaelClerx:
Looks good! Left some comments and questions, but seems a sensible approach as far as I can see just now...

fc/code_generation.py
jonc125: Like I said, it's hacky :) I'll move things around and document before asking for a re-review.

fc/protocol.py
jonc125: Yeah, it's weird, but was the closest match I could find without inventing our own property, so that's what pycml uses. 'V is a version of a state variable' is the sense of it. But could also be 'IKr is a version of a potassium current' or 'O1 is a version of an open state'.

test/test_generated_model_graphstate.py
jonc125: The check against reference results does that.

MichaelClerx: Woops. Github hid that bit from me!

jonc125:
I'm thinking about where to tidy the model manipulation stuff too, rather than a very long `set_model` method! What do you think of the following @MichaelClerx ?

* `get_variables_transitively` to new `fc.parsing.rdf` module (since its now used outside `code_generation.py`)
* Everything in `set_model` after `self.model_interface.associate_model(model, self.units)` and before the `create_weblab_model` call to a method on `fc.parsing.actions.ModelInterface`, which will call smaller methods (on that class or on other actions) to do the individual manipulations.
* `ModelInterface` can gain `outputs`, `parameters`, and `vector_orderings` properties that can be passed directly to `create_weblab_model`.

Do you want me to make the changes in this PR, or merge this and do the refactoring in a separate PR?

MichaelClerx:
Sounds good. Happy for that to go in this PR!

jonc125:
Hah, I was creating #94 just as you posted that! Anyway, merged it in here, and ready for re-review.

fc/code_generation.py
```diff
@@ -195,45 +192,47 @@ def derivative_name(deriv):
...
-                        for x in state_info]
-            parameter_index = None
-            length = len(state_info)
+        term = output.rdf_term
+        symbols = get_variables_transitively(model, term)
+        output_symbols.update(symbols)
+        if len(symbols) == 0:
+            raise ValueError('No variable annotated as {} found'.format(term))
+        elif len(symbols) == 1:
+            length = None  # Not a vector output
```
MichaelClerx: This `length = None` distinguishes between scalars and length-1 vectors, but the call to `get_variables_transitively` does not?

jonc125: You find out whether it's a scalar or vector output by seeing how many variables are annotated with the term.

jonc125: As in, there is no concept of a length-1 vector (at present).

fc/parsing/actions.py
```diff
@@ -521,47 +523,101 @@ def _expr(self):
...
 
-    ``input <prefix:term> [units <uname>] [= <initial_value>]``
     """
     def _expr(self):
         name = self.prefixed_name = self.get_named_token_as_string('name')
         self.ns_prefix, self.local_name = name.split(':', 1)
         self.ns_uri = None  # Will be set later using protocol's namespace mapping
+        return self
+
+    @property
```
MichaelClerx: This seems really hacky to me!
It's a property that will give some kind of error if I use it before the parent object is in a certain state?
Let's just make it a property that's either set or not, or a method that can raise an exception

fc/parsing/actions.py
```diff
@@ -521,47 +523,101 @@ def _expr(self):
...
+    def rdf_term(self):
+        """The RDF term annotating this variable."""
+        return create_rdf_node((self.ns_uri, self.local_name))
+
+    @classmethod
```
MichaelClerx: Any reason why this method can't live in the test code?

fc/parsing/actions.py
```diff
@@ -746,26 +805,50 @@ def resolve_namespaces(self, ns_map):
         for item in itertools.chain(self.inputs, self.outputs, self.optional_decls):
             item.ns_uri = ns_map[item.ns_prefix]
 
-    def associate_model(self, model, units):
-        """Tell this interface what model it is being used to manipulate.
+    def modify_model(self, model, units):
+        """Use the definitions in this interface to transform the provided model.
```
MichaelClerx: Maybe a quick mention or a list of the things that get modified?

fc/parsing/actions.py
jonc125: Because the test code then needs to know details about exactly what properties to set on this class, which seemed ugly to me.

fc/parsing/actions.py
jonc125: I might fill that in later, but it's given in the docstrings of the individual methods for what's already implemented, and more will be added as we go.

fc/parsing/actions.py
```diff
@@ -746,26 +805,50 @@ def resolve_namespaces(self, ns_map):
...
 
-        :param cellmlmanip.model.Model model: the model, for resolving variable references via ontology terms
+        :param cellmlmanip.model.Model model: the model to manipulate
         :param cellmlmanip.units.UnitStore units: the protocol's unit store, for resolving unit references
         """
-        self._model = model
-        self._units = units
+        self.model = model
```
MichaelClerx: Do we need to store these here? The method description suggests this can be used on several models, so would prefer to pass model & units along to any method that needs them instead of storing

fc/parsing/actions.py
MichaelClerx: You'd still have to read the code to see that the methods are called... But maybe I'm thinking too much "library" and not enough "application" here...?

fc/parsing/actions.py
jonc125: Well, all the users are in this file and it only gets accessed in the context of manipulating a model, by which point it *is* set. The docstring was just a warning if in the future it starts being used elsewhere.

fc/parsing/actions.py
jonc125: I've no objection to listing the methods called in order in the docstring. But this will probably change as we add more features, and so will easily get out of sync. I thought it'd be better to document once stabilised.

And yes, I'm treating this more as an application than a library :)

fc/parsing/actions.py
jonc125: I'm not sure we can easily pass them around, as pretty much everything needs them, including things like the symbol and number generators that are themselves passed around as functions. Since the `modify_model` method is treated as the main entry point, everything needing them will have them set correctly. But in any case all our uses at present only modify one model.

fc/parsing/actions.py
MichaelClerx: :D That doesn't make it better!

fc/parsing/actions.py
MichaelClerx: I always try and make the documentation say what the _intention_ of the code is (because you can't tell from the code whether it does what it's supposed to do, you can only see what it does).
Could we stick in something like

``The following modifications are made:
- Converting the time units if required (see :meth:...)
- ...
```
with a very obvious `...` to show this documentation isn't finished?

I know we're the only users at present but the whole goal of this project is to make modelling reliable, reproducible, and trustworthy. That means it should be extremely clear to WL users what the WL does, and they should be able to strip away complexities like the web interface and just mess around with FC locally if they are surprised at any of the results. To me the whole project is meaningless if the WL itself isn't easy to open up, inspect, and understand

Comment (jonc125) [reply fc/parsing/actions.py]: But at the point I wrote this I wasn't sure what the intent was, entirely. It's clearer in #95 so I could probably write the docstring out further there if you like?

File: fc/parsing/actions.py [hunk 10]
```diff
@@ -780,6 +864,99 @@ def sympy_equations(self):
...
+
+    def _add_input_variables(self):
+        """Ensure input variables exist in the desired units.
+
+        TODO: At present this just checks they exist; no units conversion or initial_value handling.
+        TODO: If an input doesn't exist but has units & initial_value here, or is defined via
+        an equation, then add a new variable to the model.
+        """
+        for var in self.inputs:
+            assert var.units is None
```
MichaelClerx: Why no units?

jonc125: Because the case where there are units specified isn't implemented yet.

fc/parsing/actions.py
```diff
@@ -780,6 +864,99 @@ def sympy_equations(self):
...
+                assert var.initial_value is not None  # TODO: Maybe throw ProtocolError instead?
+            else:
+                var = lhs
+                var.initial_value = None  # In case it was a state variable previously
+            # Figure out if this is a replace or add
+            defn = self.model.get_definition(var)
+            if defn is not None:
+                self.model.remove_equation(defn)
+            self.model.add_equation(eq)
+        # TODO: Check units of newly added equations; apply conversions where needed?
```
MichaelClerx: Maybe already add the check, but raise some sort of NotImplementedError? That way we'll know where to look when we start implementing :D

jonc125: I've stuck TODO comments in for everything not handled yet, so we can just search for those. Not sure yet exactly what the check should look like!

fc/parsing/actions.py
```diff
@@ -780,6 +864,99 @@ def sympy_equations(self):
...
+
+    def _annotate_state_variables(self):
+        """Annotate all state variables with the 'magic' oxmeta:state_variable term."""
+        state_annotation = create_rdf_node((OXMETA_NS, 'state_variable'))
+        self.vector_orderings[state_annotation] = {}
+        for i, state_var in enumerate(self.model.get_state_symbols()):
+            if not state_var.cmeta_id:
+                state_var.cmeta_id = self.model.get_unique_cmeta_id(state_var.name.replace('$', '__'))
```
MichaelClerx: Should we maybe get the "$" sign from cellmlmanip directly?

jonc125: https://github.com/ModellingWebLab/cellmlmanip/issues/228 suggests moving this into a cellmlmanip method anyway.

fc/parsing/rdf.py
```diff
@@ -0,0 +1,52 @@
...
+
+def get_variables_transitively(model, term):
+    """Return a list of variables annotated (directly or otherwise) with the given ontology term.
+
+    Direct annotations are those variables annotated with the term via the bqbiol:is or
+    bqbiol:isVersionOf predicates.
+
+    However we also look transitively through the 'oxmeta' ontology for terms belonging to the RDF
+    class given by ``term``, i.e. are connected to it by a path of ``rdf:type`` predicates, and
+    return variables annotated with those terms.
```
MichaelClerx: Can you give an example here?

MichaelClerx: I'm not sure how to understand "the RDF class given by ``term``"

test/test_code_generation.py
```diff
@@ -50,23 +52,31 @@ def test_generate_weblab_model(tmp_path):
...
+    vector_orderings = {state_annotation: {}}
+    for i, state_var in enumerate(model.get_state_symbols()):
+        if not state_var.cmeta_id:
+            state_var.cmeta_id = model.get_unique_cmeta_id(state_var.name.replace('$', '__'))
+        model.rdf.add((state_var.rdf_identity, PRED_IS_VERSION_OF, state_annotation))
+        vector_orderings[state_annotation][state_var.cmeta_id] = i
+
     # Create weblab model at path
-    cg.create_weblab_model(str(path), class_name, model, outputs, parameters)
+    cg.create_weblab_model(str(path), class_name, model, outputs, parameters, vector_orderings)
```
MichaelClerx: Are we testing that `vector_orderings` are actually applied here?

test/test_code_generation.py
```diff
@@ -50,23 +52,31 @@ def test_generate_weblab_model(tmp_path):
...
-    # Select model parameters (as qualified names)
+    # Select model parameters
     parameters = [
-        (oxmeta, 'membrane_fast_sodium_current_conductance'),
-        (oxmeta, 'membrane_potassium_current_conductance'),
+        VariableReference.create('oxmeta', OXMETA_NS, 'membrane_fast_sodium_current_conductance'),
+        VariableReference.create('oxmeta', OXMETA_NS, 'membrane_potassium_current_conductance'),
     ]
 
+    # Annotate state variables with the magic oxmeta:state_variable term
```
MichaelClerx: Why are we doing this? Just to test that it can deal with them already being present? Should we have a test for when they're not present or partially present?

fc/parsing/actions.py
MichaelClerx: Yes please!

fc/parsing/actions.py
MichaelClerx: Fair enough!

test/test_code_generation.py
jonc125: If they weren't, the state variable vector isn't guaranteed to be in the correct order, so that provides some level of testing. I'm not sure if it would definitely be wrong in this test, but it was in the GraphState ones.

fc/parsing/rdf.py
jonc125: OK, will try to do that.

test/test_code_generation.py
jonc125: It's needed here because it's this annotation that create_weblab_model now uses to build the state variable vector, so without it the vector would be empty.

fc/parsing/rdf.py
jonc125: Done.

MichaelClerx: Thanks!

test/test_code_generation.py
MichaelClerx: So will we need to add it in any models we use? Otherwise not sure why it's not inside of create_weblab_model ?

jonc125: The `modify_model` logic sets it normally; it's only here in the test because that machinery isn't being used. No need to add it to models. It's not inside `create_weblab_model` because setting it correctly needs info that function doesn't have, so that e.g. state variables that are units-converted as outputs are reported correctly in the `get_outputs` method of the generated model.

PR #98, opened 2020-02-14
Implement simple clamping

### LINKED ISSUE #96: Run test_clamping 1 & 2 protocols
Part of #23.

- [x] Should be just a matter of implementing the `ClampVariable` action to delegate to `ModelEquation`, providing syntactic sugar.

Tested in `test_algebraic_models.py`. Which means we also need:
- [x] Handle models with no free variable in code generation.

- [x] The `var = var` equation that clamping to initial value results in gets simplified to just `BooleanTrue` by sympy. We might need to represent this differently in sympy, and/or handle it specially in `_add_or_replace_equations`. I suspect the best option is only to treat clamping to fixed value as syntactic sugar, and keep a separate action for clamping to initial value. All such actions can then be processed separately just before or after `_add_or_replace_equations`.
    - Question: if the RHS of the existing equation defining `var` refers to other variables, should this be allowed?

Possible extension:
- [ ] Check that a variable isn't defined more than once in the interface. Have a check before making changes that we don't have 2 define and/or clamp clauses targeting the same variable, and produce a friendly error.

jonc125:
Fixes #96, at least the core functionality needed for the clamping test protocols.


jonc125:
I'm not sure why Travis isn't triggering for PRs that target another branch :(

jonc125:
> I'm not sure why Travis isn't triggering for PRs that target another branch :(

Figured it out!

.travis.yml
```diff
@@ -6,10 +6,6 @@ addons:
   apt:
     packages:
       - libsundials-serial-dev
-branches:
```
MichaelClerx: I think this was here to prevent PRs from being built twice (once as branch, once as PR)

fc/parsing/actions.py
```diff
@@ -900,7 +961,21 @@ def _add_or_replace_equations(self):
...
-        # TODO: Check units of newly added equations; apply conversions where needed?
+        # TODO: Check units of newly added equations; apply conversions where needed now or later?
+
+    def _handle_clamping(self):
+        """Clamp requested variables to their initial values."""
+        for clamp in self._clamps:
+            var = self.model.get_symbol_by_ontology_term(clamp.rdf_term)
+            defn = self.model.get_definition(var)
+            if var.initial_value is None:
+                value = defn.rhs.evalf()  # TODO: What if there are variable references in the RHS?
```
MichaelClerx: Is this for a future test case? Will evalf() fail here at least?

fc/parsing/actions.py
```diff
@@ -956,6 +1034,9 @@ def _purge_unused_mathematics(self, output_symbols):
         # Remove them and their definitions
         for symbol in unused_symbols:
             self.model.remove_variable(symbol)
+        # Add time back in to the graph if needed
+        if time_symbol not in self.model.graph.nodes:
```
MichaelClerx: Why would it have disappeared?

jonc125: If the model modifications make the model algebraic (i.e. no ODEs) - which is the case for the clamping protocols.

fc/parsing/actions.py
jonc125: Yes, I think evalf would fail in this case. It was the question I posed on #96.

.travis.yml
jonc125: Yeah, unfortunately it has an undesirable side effect! We can turn off branch builds in the Travis web config, but then you don't end up with builds of master for readme badges etc I think. Or we could turn off normal branch builds and have a cron build of master each day?

fc/parsing/actions.py
MichaelClerx: Interesting! How does that happen? The remove_variable call above shouldn't ever remove the time variable, and I don't see any remove_variable calls in cellmlmanip itself?

jonc125: It's not that the variable is removed, just that it doesn't appear in model.graph as it isn't referenced in any equations.

.travis.yml
MichaelClerx: Hmmm... The `only: -master` is meant to make master build anyway. We use the same config without problems in PINTS

jonc125: Yeah, having that config worked for that. But what it broke was PRs only got built if they were against master, not if they were building on top of another branch.

fc/parsing/actions.py
MichaelClerx: Oh right, cause the graph gets rebuilt after the manipulations. Wonder if we should fix it at cellmlmanip level that the time variable is in there?

.travis.yml
MichaelClerx: Oh yeah. I've seen that behaviour too!

fc/parsing/actions.py
jonc125: I wondered about that. It is of course possible for a CellML file not to have time at all! And if it didn't have any ODEs in the maths, I'm not sure how you'd tell what to consider time to be?

jonc125: Whereas here I know, because it was either in the original model or the protocol created it.

PR #100, opened 2020-02-17
Stop test_output_folder creating testoutput folder

### LINKED ISSUE #15: Move `testoutput` into `test` directory
Or even better, use a `tempdir`

jonc125:
Fixes #15. Part of #2.

Note that some tests currently fail because this branch hasn't adapted to the latest cellmlmanip changes - will be fixed by #93.

### DIFF
--- a/fc/file_handling.py
+++ b/fc/file_handling.py
@@ -7,6 +7,9 @@ from .error_handling import ProtocolError
 
 class OutputFolder(object):
     """
+    Manages creation and safe clean-up of folders for protocol output to be written to, in a central location defined by
+    the environment variable ``CHASTE_TEST_OUTPUT``.
+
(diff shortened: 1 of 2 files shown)


MichaelClerx:
What's the point of the `OutputFolder` class? 

> This class contains routines providing the key functionality of the C++ classes OutputFileHandler and FileFinder.
> In particular, it allows the creation of output folders within the location set for Chaste output, and safe deletion of such folders, without risk of wiping a user's drive due to coding error or dodgy path settings.

This doesn't really explain what it does or why we need this outside of Chaste

jonc125:
It manages creating folders for protocol output to be written to, in a central location defined by environment variable (if not specified otherwise by the user). And cleaning up such folders automatically.

Issue #95, opened 2020-02-14
Run test_sim_environments protocol


jonc125:
Requires support for:
- [x] Model input that specifies initial_value for state variable (V)
- [x] Model input that doesn't exist in the model (SR_leak_current), creating the variable
- [x] ~Model input that specifies units, with conversion if necessary - only if we also test on a model that has SR_leak_current, need to check what would be suitable~
- [x] A `define` clause with a variable on the RHS - should be handled by existing code
- [x] A model parameter (SR_leak_current)
    * Need to think about how we want to handle setting the initial_value for this - possibly set the `VariableDummy` attribute? Because we'll never actually generate an equation setting it; it'll be read/written in the parameters vector.

jonc125:
I've now sketched out how to tackle this, and some of #97 & #99, in https://github.com/ModellingWebLab/weblab-fc/compare/96-simple-clamping...95-sim-environments. Probably won't have time to get back to it until Thu/Fri, so feel free to pick up anyone if you get there before me!

MichaelClerx:
@jonc125 re:

> only if we also test on a model that has SR_leak_current, need to check what would be suitable

did you mean test on a model that does _not_ have SR_leak_current? (In that case we get unit conversion from V to mV with paci. the SR leak is set to something not compatible with current so should always fail in a sensible model.)

jonc125:
I don't think we need this with your new test cases.

Issue #23, opened 2019-02-18
Run further test protocols


MichaelClerx:
(Jonathan says:)

Basically here we want to add features iteratively, using existing protocols to test they're working as much as possible. I've tried to suggest a sensible ordering below. These list the protocol used; the actual Python tests using them may be named differently... Note also that some were only run in C++ tests; see https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/FunctionalCuration/test

- [x] GraphState includes simple units definitions, `independent var units ...`, `output ...` (including for the state vector), and `output ... units ...` #22 
- [x] test_while_loop adds `input ...` and overriding an existing equation #30 
- [x] test_sim_environments checks creation/overriding of input variables and replacing an equation #95
- [x] test_clamping1 adds `clamp ...`; test_clamping2 adds `clamp ... to ...` #96
    - [x] test_clamping3 adds the `units_of` construct (could be done later) #106
- [x] test_mathml_evaluation includes `output ... units ...` (including for variables created by the protocol), `input ... units ... = ...`, `define ... = ...`, and tests correct handling of (most of) the mathematics it's possible to define in a protocol #97
- [x] test_INa_IV_curves has some more complex units definitions, `optional ...`, `var ... units ... = ...`, and `convert ... to ... by ...`; similarly for TestICalProto.py #99
- [x] test_other_ontologies has terms not in `oxmeta` and a generic vector output (not the state vector) - should be covered by #95 and #97 actually - but doesn't seem to pass without further changes, so made ticket: #141
- [x] test_S1S2 checks import handling and redefining the stimulus current #106
- [x] test_nested_protocol makes sure the model interface from a nested protocol is handled correctly, and also includes `define diff(...) = ...` #107
    - [x] The _block protocols test merging model interface sections between parent & nested protocols, where both have parts of the interface defined #107
- [x] test_original_definition checks `original_definition` #108
- [ ] test_model_interpolation has several tests of the `interpolate()` special function #109

MichaelClerx:
Only #109 left now, so might as well close this ticket as a duplicate of that one :D

PR #103, opened 2020-02-19
Preparatory work and documentation

jonc125:
Part of #95, but I suspect it would be worth merging in this state since it documents where other parts of #23 might need to make changes! Then actually implementing #95 could be a separate PR.

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -11,12 +11,13 @@ import sympy
 from cellmlmanip.model import DataDirectionFlow
 from cellmlmanip.parser import UNIT_PREFIXES
 
+from ..error_handling import ProtocolError
 from ..language import expressions as E
 from ..language import statements as S
 from ..language import values as V
 from ..locatable import Locatable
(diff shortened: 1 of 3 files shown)


fc/parsing/actions.py
```diff
@@ -819,16 +820,36 @@ def resolve_namespaces(self, ns_map):
...
         """
         self.model = model
         self.units = units
         self._convert_time_if_needed()
         self._add_input_variables()
+        # TODO: Add variables defined by DeclareVariable
         self._add_or_replace_equations()
         self._handle_clamping()
         self._annotate_state_variables()
+        # TODO: Resolve initial_values on non-states, maybe using Model.transform_constants
```
MichaelClerx: Not sure what you mean by resolving them? Cellmlmanip should already have converted those to equations

jonc125: It will for ones in the model, but not necessarily for variables added/changed by the protocol, i.e. inputs.

PR #113, opened 2020-02-24
Updated code to work with recent cellmlmanip changes.

MichaelClerx:
https://github.com/ModellingWebLab/cellmlmanip/pull/229

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -982,10 +982,10 @@ class ModelInterface(BaseGroupAction):
         state_annotation = create_rdf_node((OXMETA_NS, 'state_variable'))
         self.vector_orderings[state_annotation] = {}
         for i, state_var in enumerate(self.model.get_state_symbols()):
-            if not state_var.cmeta_id:
-                state_var.cmeta_id = self.model.get_unique_cmeta_id(state_var.name.replace('$', '__'))
+            if not state_var.rdf_identity:
+                self.model.add_cmeta_id(state_var)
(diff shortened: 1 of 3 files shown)


fc/parsing/actions.py
```diff
@@ -982,10 +982,10 @@ def _annotate_state_variables(self):
         state_annotation = create_rdf_node((OXMETA_NS, 'state_variable'))
         self.vector_orderings[state_annotation] = {}
         for i, state_var in enumerate(self.model.get_state_symbols()):
-            if not state_var.cmeta_id:
-                state_var.cmeta_id = self.model.get_unique_cmeta_id(state_var.name.replace('$', '__'))
+            if not state_var.rdf_identity:
```
jonc125: We can remove the `if` here, and just do `self.model.add_cmeta_id(state_var)` since it's a no-op if one exists.

test/test_code_generation.py
```diff
@@ -70,10 +70,10 @@ def test_generate_weblab_model(tmp_path):
     state_annotation = create_rdf_node((OXMETA_NS, 'state_variable'))
     vector_orderings = {state_annotation: {}}
     for i, state_var in enumerate(model.get_state_symbols()):
-        if not state_var.cmeta_id:
-            state_var.cmeta_id = model.get_unique_cmeta_id(state_var.name.replace('$', '__'))
+        if not state_var.rdf_identity:
```
jonc125: Same here.

PR #115, opened 2020-02-25
Adapted code for recent cellmlmanip changes.

MichaelClerx:
See https://github.com/ModellingWebLab/cellmlmanip/pull/237


MichaelClerx:
This fails since the cellmlmanip PR hasn't been merged, but tested locally and works!

MichaelClerx:
Got the OK from Sarah!

PR #117, opened 2020-02-25
Set initial values defined in `inputs`

MichaelClerx:
Part of #95


MichaelClerx:
@jonc125 could you have an early look at the way inputs and defines are handled here?

(As a complete distraction: It currently fails because the unique name generating code expects every variable to have a component in its name! Will rewrite that method tomorrow)

fc/parsing/actions.py
```diff
@@ -921,6 +930,9 @@ def _add_input_variables(self):
         units are given. If not, an error is raised.
         """
         for var in self.inputs:
+
+            # Find variable symbol or create a new one
+            symbol = None
```
jonc125: I think you want `variable = None` here?

fc/parsing/actions.py
```diff
@@ -934,32 +946,66 @@ def _add_input_variables(self):
...
+        Process define statements and modify the model's equations accordingly. Also sets initial values determined in
+        input statements.
+        """
+        # Apply all modifications (sympy_equations contains _only_ equations from define statements)
         for eq in self.sympy_equations:
             lhs = eq.lhs
             if lhs.is_Derivative:
                 var = lhs.args[0]
-                if var.initial_value is None:
+                term = self.model.get_ontology_terms_by_variable(var)[0]
```
jonc125: This isn't guaranteed to work. Instead I'd do a pre-processing phase before you iterate through equations, where instead of a map from RDF term to initial value you create a map from variable to initial value, looking up the terms. Then you can look up directly in that map here.

jonc125: Otherwise this looks good!

fc/parsing/actions.py
```diff
@@ -934,32 +946,66 @@ def _add_input_variables(self):
...
+            # Remove existing equation if needed
+            old_eq = self.model.get_definition(var)
+            if old_eq is not None:
+                self.model.remove_equation(old_eq)
+
+            # Add new equation
             self.model.add_equation(eq)
+
+        # Apply all initial values set in the inputs
+        for term, value in self.initial_values.items():
```
jonc125: And you can re-use that pre-processed map here.

jonc125: And you can re-use that pre-processed map here.

fc/parsing/actions.py
```diff
@@ -953,7 +952,7 @@ def _add_input_variables(self):
 
             # Store initial value if given
             if variable is not None and var.initial_value is not None:
-                self.initial_values[var.rdf_term] = var.initial_value
+                self.initial_values[variable] = var.initial_value
```
jonc125: I didn't mean to resolve annotations here, since later units conversions might move the annotation. In general we only want to resolve annotations to variables at the point of need.

fc/parsing/actions.py
```diff
@@ -965,19 +964,18 @@ def _add_or_replace_equations(self):
             lhs = eq.lhs
             if lhs.is_Derivative:
                 var = lhs.args[0]
-                term = self.model.get_ontology_terms_by_variable(var)[0]
 
                 # Initial value must be set via inputs, or already be set (if this was already a state variable)
-                if var.initial_value is None and term not in initial_values:
+                if var.initial_value is None and var not in initial_values:
+                    term = self.model.get_ontology_terms_by_variable(var)[0]
```
jonc125: When looping over self.initial_values to create the var -> init map, you could also create a var -> term map to ensure the term the user specified is given in the error message.

fc/parsing/actions.py
MichaelClerx: Hmmm. Why is unit conversion only done at the very end of `modify_model` ?
(I can see why it's not done at this point yet, but why do it after handling defines and clamps?)

jonc125: In pycml it was done at the end, because that was how the architecture was set up - it was too complex to do it bit by bit as you went along. That might no longer be the case with cellmlmanip - and indeed we're converting time at the start. But I've not fully thought through all the details.

Regardless, it's safer only to resolve annotations to variables when needed as then you don't have to worry about other manipulations invalidating your map, and hence getting unexpected side effects.

MichaelClerx: But... how can I set an initial value or replace an equation before units have been converted? Surely the new definition is in new units, while the variable still reports as being in the old units?

jonc125: The later pass by pycml fixed that up, putting a conversion in where the LHS & RHS of equations it had added/changed reported different units. I'm not sure if we want to do the same in weblab-fc or not. But that doesn't affect my point about the change here!

MichaelClerx: Ok, but I have no idea what to do anymore

jonc125: Sorry, was my initial suggestion unclear? All I meant was to keep this line as `self.initial_values[var.rdf_term] = var.initial_value` and create another map from `self.initial_values` *within* the `_add_or_replace_equations` (as the first thing it does) that does variable -> value, looking up variables by ontology term at that point.

I don't think we have a test case for units converting inputs yet - when we do we'll need to figure that out. But that's independent from getting this merged so #95 works in the matching units case.

fc/parsing/actions.py
jonc125: The reason being, a variable could be annotated by multiple different terms, possibly from different ontologies.

MichaelClerx: But the term is only used for variables that are _not_ in `self.initial_values`!

MichaelClerx:
@jonc125 shall I keep working on other bits of #95 here? Or do a separate PR?

jonc125:
I'm happy to keep going on this PR, but we can merge and start a new one if you prefer. Looks like some fixes are needed for Maurice's change to cellmlmanip too.

MichaelClerx:
Let's keep going then!

> Looks like some fixes are needed for Maurice's change to cellmlmanip too.

Do you think this PR will be affected by https://github.com/ModellingWebLab/cellmlmanip/pull/253 ?

fc/parsing/actions.py
jonc125: Good point!

jonc125:
> Do you think this PR will be affected by [ModellingWebLab/cellmlmanip#253](https://github.com/ModellingWebLab/cellmlmanip/pull/253) ?

I'm hoping not...

MichaelClerx:
I could probably stop renaming the tests as "_generated_model_" now :D

MichaelClerx:
Appears to pass tests now @jonc125 

Although we didn't hit this point

> - [ ] Model input that specifies units, with conversion if necessary - only if we also test on a model that has SR_leak_current, need to check what would be suitable

?

fc/parsing/actions.py
```diff
@@ -1002,6 +1060,21 @@ def _convert_output_units(self):
...
         return output_variables
 
+    def _list_parameters(self):
+        """Populates the list of model parameters (inputs with a constant RHS)."""
+
+        for var in self.inputs:
+            try:
+                variable = self.model.get_variable_by_ontology_term(var.rdf_term)
+            except KeyError:
+                # Skip optional variables
```
jonc125: I think until this is implemented properly, we should be raising rather than skipping here.

test/test_generated_model_sim_environments.py
```diff
@@ -1,11 +1,8 @@
 import os
```
jonc125: I thought you weren't going to rename tests any more???

fc/parsing/actions.py
```diff
@@ -1002,6 +1060,21 @@ def _convert_output_units(self):
...
+                continue
+
+            # Parameters are inputs that aren't states, and have a constant RHS
+            eq = self.model.get_definition(variable)
+            if isinstance(eq.lhs, VariableDummy) and len(eq.rhs.atoms(VariableDummy)) == 0:
```
jonc125: Not something to do in this PR, but perhaps worth discussing with @MauriceHendrix whether a `Model.is_parameter(self, variable)` method that does this in cellmlmanip would be useful?

test/test_generated_model_sim_environments.py
MichaelClerx: I needed a change to rebuild :D
Have given all 3 their original names now

fc/parsing/actions.py
MichaelClerx: In this case I think the skip is what we'll want? The code in `_add_input_variables` could do with raising something maybe?

jonc125: Possibly. I guess we'll find out for sure when we implement optional handling!

Issue #112, opened 2020-02-24
Make tests run faster


MichaelClerx:
- [x] Run profiling to see where most time is spent
- [ ] Presumably, fix #41

Current tests take 10-15 minutes and use excessive computing power

MichaelClerx:
Added [pytest-profiling](https://pypi.org/project/pytest-profiling/) plugin (see contributing.md).

Not sure how good it is, doesn't seem to filter out all of pytest's calls

MichaelClerx:
Example SVG file generated for `test_algebraic_models.py`

[combined.zip](https://github.com/ModellingWebLab/weblab-fc/files/4250729/combined.zip)

Seems to indicate that all time is spent parsing, rather than manipulating or running experiments @jonc125 ?

jonc125:
That would make sense; the protocols themselves are tiny so running will be small. Parsing the model is ~3%, compiling Cython code ~7.5%, modifying ~2%. pyparsing 77% and converting the parse actions via expr() 10%.

MichaelClerx:
On my laptop, the time taken for `test_algebraic_models` (without profiling enabled) is `87.8s`, so that means `68` seconds spent in `pyparsing`

MichaelClerx:
So 34 seconds per protocol. Persumably this is just time spent parsing the library files? It can't take 34 seconds to parse those protocol files themselves, can it?

jonc125:
If they're now fast enough to be live-able with, then yes. @skeating ?

PR #119, opened 2020-02-25
Some investigation of parsing speed behaviour

jonc125:
Some investigating around #112. Run tests with `-s` flag to see the pyparsing debug output.

Bizarrely, with `time pytest test/test_algebraic_models.py -s` the second protocol always throws a recursion error, even though it's almost identical to the first and the cache has been reset. So there seems to be something a little screwy going on with shared state: this doesn't happen if it's the only test you run (with an extra `-k init`).

The packrat cache seems to be hit about as often as it's missed, which suggests that it is having some benefit. I discovered when trying to increase its size that matplotlib enables it before we get there, so have added a hacky re-enable in!

Feel free to have a play @MichaelClerx :)

https://stackoverflow.com/questions/51412184/pyparsing-packrat-slows-down-performance may be worth a read.

### DIFF
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -9,7 +9,10 @@ from . import actions
 __all__ = ['CompactSyntaxParser']
 
 # Necessary for reasonable speed when using infixNotation
-p.ParserElement.enablePackrat()
+# Also, we force re-enable packrat, because matplotlib can have done it already with different settings!
+p.ParserElement._packratEnabled = False
+p.ParserElement.resetCache()
+p.ParserElement.enablePackrat(None)
 
 
 ################################################################################
@@ -583,7 +586,8 @@ class CompactSyntaxParser(object):
             raise RuntimeError("Failed to parse expression even with a recursion limit of %d; giving up!"
                                % (int(self._stack_depth_factor * self._original_stack_limit),))
         actions.source_file = orig_source_file
-        print('Packrat stats: hits={} misses={}'.format(*p.ParserElement.packrat_cache_stats), file=sys.stderr)
+        print('Packrat stats: len={}, hits={}, misses={}'.format(
+            p.ParserElement.packrat_cache.__len__(), *p.ParserElement.packrat_cache_stats), file=sys.stderr)
         p.ParserElement.resetCache()
         return r
 
--- a/fc/parsing/CompactSyntaxParser.py
+++ b/fc/parsing/CompactSyntaxParser.py
@@ -547,7 +547,8 @@ class CompactSyntaxParser(object):
         # basis that if one expression needs to, several are likely to.
         self._stack_depth_factor = 1
         self._original_stack_limit = sys.getrecursionlimit()
-        self.increase_stack_depth_limit()
+        print('Original stack limit:', self._original_stack_limit, file=sys.stderr)
+        self.increase_stack_depth_limit(step=1.0)
 
     def __del__(self):
         """Reset the stack limit if it changed."""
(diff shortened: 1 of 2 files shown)


MichaelClerx:
I don't really know anything about this, but happy to have a look!

As far as I remember, at the start of the WL project you thought the bottleneck was library parsing and were thinking to start caching it?

jonc125:
It's almost certainly parsing expressions with infixNotation that's the killer, as that will generate hugely recursive parse behaviour. Since most protocols parse the basic library (and some the cardiac library), if that could be done as a pre-processing step somehow and shared (maybe across the whole test session?) you'd probably gain significantly.

Possible approaches:
1. Have a global cache of protocol file -> parse action tree for imported protocols, and make `Protocol._parse` just use this if possible, and add to it if not.
2. Pickle the parse action tree and save it in the `fc` source, so if you're asked to parse a file that has a pickle alongside you just reload. Depends on the actions being pickle-able, and checking whether the action module has changed to make the pickle invalid, etc. so possibly risky? But would help with the typical Web Lab use-case of a single protocol being parsed, which the first approach wouldn't.

jonc125:
It's almost certainly parsing expressions with infixNotation that's the killer, as that will generate hugely recursive parse behaviour. Since most protocols parse the basic library (and some the cardiac library), if that could be done as a pre-processing step somehow and shared (maybe across the whole test session?) you'd probably gain significantly.

Possible approaches:
1. Have a global cache of protocol file -> parse action tree for imported protocols, and make `Protocol._parse` just use this if possible, and add to it if not.
2. Pickle the parse action tree and save it in the `fc` source, so if you're asked to parse a file that has a pickle alongside you just reload. Depends on the actions being pickle-able, and checking whether the action module has changed to make the pickle invalid, etc. so possibly risky? But would help with the typical Web Lab use-case of a single protocol being parsed, which the first approach wouldn't.

MichaelClerx:
Two sounds best then.

I don't suppose pyparsing has any options for non-recursive parsing of equations? (with queues and stacks and such)

MichaelClerx:
Actually, those libraries don't even look very big. Hardly any long equations!
Is pyparsing just really slow?

jonc125:
No, pyparsing is a recursive descent parser; we'd need to look at other parsing libraries to avoid it. It's not so much the big expressions, as the complex expression grammar - lots of different atoms and operators, which results in lots of backtracking.

MichaelClerx:
I wonder what the best way to get some insight into this is! What have you been doing?

Maybe we could just run a debugger and start looking what it's doing about halfway through the parse?

I've been trying to compare it to other python parsing things. It takes me .75 seconds to parse the Heijman 2011 model, which is 110kb of almost pure equations, parsed with a handwritten (not very good?) recursive pure python parser. The libraries are about 26kb and take about .34s in our system. So ignoring the fact that most of the library code is comments, a speed-up of a factor 2 should be in there somehow?

Do you think there's some fiddly changes (e.g. replacing something with a regex here and there, re-ordering some operations) that would make it e.g. backtrack less?

jonc125:
It's not easy!  You can use something like this to output huge outputs of tracing of how the parser is handling the protocols (and you can target it to particular parts of the parser if you wish):
```diff --git i/test/test_algebraic_models.py w/test/test_algebraic_models.py
--- i/test/test_algebraic_models.py
+++ w/test/test_algebraic_models.py
@@ -2,6 +2,10 @@
 import unittest

 import fc
+import fc.parsing.CompactSyntaxParser as csp
+
+
+csp.enable_debug()


 class TestAlgebraicModels(unittest.TestCase):
```

I've tried in the past using regex, expression ordering, etc. so there's quite a bit of that in there already.

PR #121, opened 2020-02-28
Fix for recent changes in cellmlmanip

MichaelClerx:
https://github.com/ModellingWebLab/cellmlmanip/pull/251

### DIFF
--- a/fc/code_generation.py
+++ b/fc/code_generation.py
@@ -7,16 +7,15 @@ import time
 import jinja2
 import sympy
 
-from cellmlmanip import transpiler
 from cellmlmanip.printer import Printer
+from cellmlmanip.transpiler import Transpiler
 
 from .parsing.rdf import OXMETA_NS, get_variables_transitively
 
 # Add an `_exp` method to sympy, and tell cellmlmanip to create _exp objects instead of exp objects.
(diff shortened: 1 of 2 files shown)


fc/code_generation.py
```diff
@@ -7,16 +7,15 @@
...
+from cellmlmanip.transpiler import Transpiler
 
 from .parsing.rdf import OXMETA_NS, get_variables_transitively
 
 # Add an `_exp` method to sympy, and tell cellmlmanip to create _exp objects instead of exp objects.
 # This prevents Sympy doing simplification (or canonicalisation) resulting in weird errors with exps in some cardiac
 # models.
-setattr(sympy, '_exp', sympy.Function('_exp'))
-transpiler.SIMPLE_MATHML_TO_SYMPY_NAMES['exp'] = '_exp'
+Transpiler.set_mathml_handler('_exp', sympy.Function('_exp'))
```
jonc125: Doesn't the first argument need to be just `'exp'` as that's the MathML operator name?

jonc125: The comment above should probably be updated too!

MichaelClerx:
Another look @jonc125 ?

PR #127, opened 2020-03-04
Added test for mem-caching of protocols leading to shared ModelInterface state.

MichaelClerx:
- [x] Added test that brings out issue
- [x] Added fix


MichaelClerx:
Ready for re-review @jonc125 !

Issue #130, opened 2020-03-05
Model interface: var specified as input & output must have same units


MichaelClerx:
Currently not checked

MichaelClerx:
Added a test in https://github.com/ModellingWebLab/weblab-fc/pull/131/files

jonc125:
At least, this is the case in pycml because the implementation there makes it fiendish to allow otherwise. I think for the sake of a user's sanity it's probably a reasonable restriction anyway, but happy to hear arguments the other way if you like!

MichaelClerx:
My only objection is that the syntax makes it look like you can specify different units, whereas in fact it's 3 statements rolled into 2 constructs (x is in mV, x is an input, x is an output)

jonc125:
Indeed. So, let's not implement a specific check for this yet and see whether our natural approach to unit conversions lets it work with different units. And if it does, we can say that's a feature not a bug :)

MichaelClerx:
I'm not sure it follows unthinkingly...

    x = 1 mV
        oxmeta: membrane_voltage

add an output in V to get

    new_x = x * 1e-3
        oxmeta: membrane_voltage
    x = 1 mV

add an input in kV to get... an error? Can't assign a value to it now
Otherwise, would work out as

    new_new_x = x * 1e-3 * 1e-3
    new_x = new_new_x / 1e-3
    x = 1 mV

???

MichaelClerx:
So we'd have to do inputs first, at least, I think

jonc125:
Though we do do inputs first with the flow at present.

MichaelClerx:
Yeah just saying we can't do it without thinking it through, which would have been nice :D

MichaelClerx:
Input first case:

    x = 1 mV
        oxmeta: membrane_voltage

add an input in V to get

    new_x = 0.001 V
        oxmeta: membrane_voltage
    x = new_x * 1000 [mV/V]

add an output in kV to get

    new_new_x = new_x * 0.001 [kV/V]
        oxmeta: membrane_voltage
    new_x = 0.001 V
    x = new_x * 1000 [mV/V]

jonc125:
Actually, I've just realised why there is a problem with this: you can't annotate both the input and the output variable with the ontology term if they're different variables!

MichaelClerx:
Closed in #131

PR #131, opened 2020-03-05
Check var has same units as input and as output

MichaelClerx:
See #130

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -933,6 +933,9 @@ class ModelInterface(BaseGroupAction):
                 raise ProtocolError(
                     'The variable ' + str(var.rdf_term) + ' appears as input and output, but with different units.')
 
+        # TODO Check defines
+        # TODO Check clamps
+
     def _variable_generator(self, name):
         """Resolve a name reference within a model interface equation to a variable in the model.
 
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -837,7 +837,7 @@ class ModelInterface(BaseGroupAction):
         add_unique(self.outputs, interface.outputs)
         add_unique(self.optional_decls, interface.optional_decls)
         add_unique(self.equations, interface.equations)
-        add_unique(self.clamps, interface._clamps)
+        add_unique(self.clamps, interface.clamps)
 
         # need to be careful with time units
         # add from nested protocol if there are no time units in outer protocol
@@ -898,15 +898,16 @@ class ModelInterface(BaseGroupAction):
             # TODO: Look for variable annotated as time instead?
             raise ProtocolError('Model must contain at least one ODE.')
 
-        # Perform initial sanity checks
-        self._sanity_check()
-
         # Convert free variable units
         self._convert_time_if_needed()
 
         # Ensure inputs exist, adding them if needed, and storing initial values set by user.
         # This also performs unit conversion where needed for existing input variables.
         self._add_input_variables()
+
+        # Now that all variables are in place, perform sanity checks on model-protocol combination before making any
+        # modifications
+        self._sanity_check()
 
         # TODO: Add variables defined with ``var`` statements
 
@@ -937,14 +938,20 @@ class ModelInterface(BaseGroupAction):
         output_units = {}           # To check input vs output units
         initial_values = set()      # To check against overdefinedness
         for ref in self.inputs:
-            var = self.model.get_variable_by_ontology_term(ref.rdf_term)
+            try:
+                var = self.model.get_variable_by_ontology_term(ref.rdf_term)
+            except KeyError:
+                continue
             if var in input_units:
                 raise ProtocolError('The variable ' + str(var) + ' was specified as an input twice.')
             input_units[var] = ref.units
             if ref.initial_value is not None:
                 initial_values.add(var)
         for ref in self.outputs:
-            var = self.model.get_variable_by_ontology_term(ref.rdf_term)
+            try:
+                var = self.model.get_variable_by_ontology_term(ref.rdf_term)
+            except KeyError:
+                continue
             if var in output_units:
                 raise ProtocolError('The variable ' + str(var) + ' was specified as an output twice.')
             output_units[var] = ref.units
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -915,7 +915,7 @@ class ModelInterface(BaseGroupAction):
         # and setting any initial values defined through inputs.
         self._add_or_replace_equations()
 
-        # Process ``clamp`` statements
+        # Process ``clamp`` statements with an RHS
         self._handle_clamping()
 
         self._annotate_state_variables()
@@ -930,7 +930,15 @@ class ModelInterface(BaseGroupAction):
         # TODO: Any final consistency checks on the model?
 
     def _sanity_check(self):
-        """Perform initial sanity checks on the protocol interface."""
+        """
+        Performs some initial sanity checks on the (protocol interface + model) combination.
+
+        - Variables can only appear as an input once.
+        - Variables can only appear as an output once.
+        - Variables appearing as input and output must have the same units (if set).
+        - Variables can only bet set by a single clamp statement or define.
+
+        """
 
         # Variables can only appear as input or output once
         # If a variable appears as an input and an ouput, both must have the same units
@@ -943,7 +951,8 @@ class ModelInterface(BaseGroupAction):
             except KeyError:
                 continue
             if var in input_units:
-                raise ProtocolError('The variable ' + str(var) + ' was specified as an input twice.')
+                raise ProtocolError(
+                    f'The variable {var} (referenced by {ref.rdf_term}) was specified as an input twice.')
             input_units[var] = ref.units
             if ref.initial_value is not None:
                 initial_values.add(var)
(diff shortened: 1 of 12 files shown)


MichaelClerx:
Ready for review @jonc125 ! 

Would like to handle defines too, but not sure where they are. Is that the `DeclareVariable` class?

MichaelClerx:
Maybe that can best wait until we have new variables using `define` sorted

fc/parsing/actions.py
```diff
@@ -910,6 +913,29 @@ def modify_model(self, model, units):
...
+            if var.rdf_term in output_units:
+                raise ProtocolError('The variable ' + str(var.rdf_term) + ' was specified as an output twice.')
+            output_units[var.rdf_term] = var.units
+            units = input_units.get(var.rdf_term)
+            if var.units is not None and units is not None and var.units != units:
+                raise ProtocolError(
+                    'The variable ' + str(var.rdf_term) + ' appears as input and output, but with different units.')
+
+        # TODO Check defines
+        # TODO Check clamps
```
jonc125: Clamps could be done now. Defines aren't implemented yet (but will be `DeclareVariable`).

test/test_model_interface.py
```diff
@@ -0,0 +1,41 @@
...
+        proto.set_model('test/models/single_ode.cellml')
+        proto.run()
+
+
+def test_duplicate_output():
+    # Tests if an error is raised when a variable is specified as output twice.
+
+    proto_file = 'test/protocols/test_bad_interface_duplicate_output.txt'
+    proto = fc.Protocol(proto_file)
+    proto.set_output_folder('test_invalid_interface')
```
jonc125: This should probably be different for each test, as general best practice.

test/test_model_interface.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_duplicate_input():
+    # Tests if an error is raised when a variable is specified as input twice.
+
+    proto_file = 'test/protocols/test_bad_interface_duplicate_input.txt'
+    proto = fc.Protocol(proto_file)
+    proto.set_output_folder('test_invalid_interface')
+    with pytest.raises(ProtocolError, match='as an input twice'):
+        proto.set_model('test/models/single_ode.cellml')
+        proto.run()
```
jonc125: Presumably we never get to the `run()` part, so could cut this line?

MichaelClerx: Yeah! I wasn't quite sure if it wasn't cleaner this way, because it's not obvious from the outside which method should raise the error?

test/test_model_interface.py
MichaelClerx: Thanks!

test/test_model_interface.py
jonc125: Ah, I see what you mean. Fair enough.

MichaelClerx:
@jonc125 not sure what's going on here, I added a `clamp` but `self._clamps` stays empty?

test/protocols/test_bad_interface_clamp_and_init.txt
```diff
@@ -0,0 +1,19 @@
...
+
+namespace oxmeta = "https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#"
+
+units {
+    V = volt
+}
+
+model interface {
+    input oxmeta:membrane_voltage units V = 1
+    clamp oxmeta:membrane_voltage to 2 :: V
```
jonc125: This form is just syntactic sugar for the `define` case above. You want just `clamp oxmeta:membrane_voltage` to have something appear in `self._clamps`. And in that case I think it would be OK to have an initial value on the input.

MichaelClerx: Oh dear. So we'd have to catch duplicates at the parse level then?

MichaelClerx: Well, if we wanted to know what the user did... Less-specific error message if we catch it later

jonc125: Less-specific error is fine for now - it would appear as a straightforward over-defined variable, and should be easy for them to find.

jonc125:
This currently requires the time units to be set with `set independent var units` - it'll break if they're converted implicitly by declaring time as an input and/or output.

MichaelClerx:
> This currently requires the time units to be set with `set independent var units` - it'll break if they're converted implicitly by declaring time as an input and/or output.

Could you clarify that? It gets the free variable from model.get_free_variable(), so not sure how that's related to the time units code?

jonc125:
You update `self.time_variable` in `_convert_time_if_needed`, but not in `_add_input_variables` or `_convert_output_units`.

jonc125:
Actually the output case probably doesn't matter, since the annotated variable wouldn't actually be the free variable (just set from it). But the input case could alter the free variable.

MichaelClerx:
Ugh. Not sure what to do anymore @jonc125 

Can't do the sanity checks on the protocol alone, as that test won't catch issues like
```
x is annotated as oxmeta:x
x is annotated as oxmeta:y
oxmeta:x in V
oxmeta:y in A
```

but trying to do it on the variables requires them all to exist, so would have to go after they get created by (1) optional input code, (2) `var` statements.

Would that make sense? Or are there other constructs that can create new variables?

jonc125:
Oh yuck! I think for now I'd just do sanity checks on the protocol alone. That kind of situation is likely to be rare. Also it gets more complex, as the example you give is probably OK if at least one of the annotations is an output (since the single variable gets split into two by model modifications!); it'd only be impossible if both were inputs.

MichaelClerx:
I ran into trouble checking via the ontology terms, because the equations (which are all I can see of defines and clamp-tos) are in terms of variables, not ontology terms. Not 100% sure why, but the variable in the equation LHS didn't seem to have any ontology terms

MichaelClerx:
> Also it gets more complex, as the example you give is probably OK if at least one of the annotations is an output (since the single variable gets split into two by model modifications!); it'd only be impossible if both were inputs.

I don't really want the system to have very complex rules. I want it to have easy rules so that the users have some chance of understanding them. If that means shutting down a few (pointless) alleys then that's fine by me

MichaelClerx:
Stuck the checks _after_ any variable creation (but before modification), and passes all tests now

fc/parsing/actions.py
```diff
@@ -1007,13 +1077,20 @@ def _add_input_variables(self):
...
 
+            # Maintain link to time variable, if needed
+            is_time = variable is self.time_variable
+
             # Convert units if needed
             if var.units is not None:
                 units = self.units.get_unit(var.units)
                 if units != variable.units:
-                    print('Converting input ' + str(var.rdf_term) + ' to units ' + str(units))
+                    # print('Converting input ' + str(var.rdf_term) + ' to units ' + str(units))
```
jonc125: Delete this line ;)

fc/parsing/actions.py
```diff
@@ -910,6 +929,58 @@ def modify_model(self, model, units):
         self._purge_unused_mathematics(output_variables)
         # TODO: Any final consistency checks on the model?
 
+    def _sanity_check(self):
+        """Perform initial sanity checks on the protocol interface."""
```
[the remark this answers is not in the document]
MichaelClerx: ok!

fc/parsing/actions.py
```diff
@@ -910,6 +929,58 @@ def modify_model(self, model, units):
...
+        input_units = {}            # To check input vs output units
+        output_units = {}           # To check input vs output units
+        initial_values = set()      # To check against overdefinedness
+        for ref in self.inputs:
+            try:
+                var = self.model.get_variable_by_ontology_term(ref.rdf_term)
+            except KeyError:
+                continue
+            if var in input_units:
+                raise ProtocolError('The variable ' + str(var) + ' was specified as an input twice.')
```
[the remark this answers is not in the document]
MichaelClerx: ok!

fc/parsing/actions.py
```diff
@@ -910,6 +929,58 @@ def modify_model(self, model, units):
...
+            if ref.units is not None and units is not None and ref.units != units:
+                raise ProtocolError(
+                    'The variable ' + str(var) + ' appears as input and output, but with different units.')
+
+        # Check against overdefinedness: Variables cannot appear as an LHS of an equation more than once (e.g. used in
+        # two defines, or used in a ``clamp x to 1`` and a define), and variables clamped to their current value can not
+        # also be defined.
+        seen = set()
+        for ref in self.clamps:
+            var = self.model.get_variable_by_ontology_term(ref.rdf_term)
```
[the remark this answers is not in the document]
MichaelClerx: done!

fc/parsing/actions.py
```diff
@@ -930,7 +930,15 @@ def modify_model(self, model, units):
...
 
     def _sanity_check(self):
-        """Perform initial sanity checks on the protocol interface."""
+        """
+        Performs some initial sanity checks on the (protocol interface + model) combination.
+
+        - Variables can only appear as an input once.
+        - Variables can only appear as an output once.
+        - Variables appearing as input and output must have the same units (if set).
+        - Variables can only bet set by a single clamp statement or define.
```
jonc125: bet -> be

Issue #107, opened 2020-02-23
Merge model interface from nested protocol


jonc125:
The contents of the interface should be supported by #30 so this one isn't actually blocked.

- [x] Test the simplest case with `test_various_protocols.py:test_nested_protocols` - only the nested protocol has any interface defined
- [x] In a separate PR, add another test that runs `protocols/IK1_block.txt` on the LR91 model to check that merging interfaces when nesting works

Part of #23.

PR #134, opened 2020-03-10
run IK1_block protocol on LR91

skeating:
Finishes #107


jonc125:
Moved this work to #159 instead.

PR #138, opened 2020-03-30
Mathml evaluation

### LINKED ISSUE #97: Do the test_mathml_evaluation protocol
Part of #23.

Following from #95, needs support for:
- [x] Non oxmeta annotations - might already work?
- [x] More mathematics in `define` clauses, i.e. write lots of `to_sympy` methods.

MichaelClerx:
Fixes #97 

Seems to pass now, but doesn't have trig functions and raises 2 warnings

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -1040,7 +1040,7 @@ class ModelInterface(BaseGroupAction):
         # Add variables created by ``define`` statements - must also be listed as output or optional input
         self._add_define_variables()
 
-        # TODO: Add variables created by ``optional`` statements (must also be listed as input or output)
+        # TODO: Add variables created by ``optional`` statements with a ``default`` clause
         # TODO: Add variables created by ``var`` statements
 
         # Now that all variables are in place, perform sanity checks on model-protocol combination before making any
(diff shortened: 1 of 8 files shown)


MichaelClerx:
Ready for review now @jonc125

fc/parsing/actions.py
```diff
@@ -1119,6 +1238,34 @@ def _add_input_variables(self):
...
+            except KeyError:
+
+                # Find variable in input or output declarations
+                try:
+                    var = variables[eq.var.name()]
+                except KeyError:
+                    continue
+
+                # Add the new variable, with ontology annotation
+                name = self.model.get_unique_name('protocol__' + var.local_name)
```
jonc125: Now these 5 lines are in two places, I'd extract a helper method!

fc/templates/weblab_model.pyx
```diff
@@ -227,6 +224,9 @@ cdef class {{ class_name }}(CvodeSolver):
             'rdf': self.env,
             'cellml': self.env,
             'vCard': self.env,
+            {%- for prefix in ns_map %}
```
jonc125: You can get rid of everything apart from this loop now.

test/test_mathml_evaluation.py
```diff
@@ -1,22 +1,18 @@
+"""Test implementation of voltage clamps.
```
jonc125: I think some dodgy copy & paste has happened here in the past!

MichaelClerx: I copied the docstring from the class to the module level. Turned it into comments now if you prefer?

jonc125: It wasn't that, it's that the docstring doesn't match what the test is doing :)

MichaelClerx: Fair enough. I forgot to read :man_facepalming:

Issue #106, opened 2020-02-23
Support simple protocol imports and units_of()


jonc125:
Test by creating a Python version of https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/FunctionalCuration/test/TestS1S2Protocol.hpp

- [x] Test `units_of` first with `test_clamping.py:test_clamping_with_variable_units` - can be separate PR
- [x] Then LR91 model tests basic import functionality
    - [x] Top-level import of a protocol, merging model interface
    - [x] Overriding stimulus current definition - should be done by #97 but check!
    - [x] The `units_of` construct in a larger equation: `0 :: units_of(oxmeta:membrane_stimulus_current_amplitude)`
- [x] Noble model tests with time units conversion - shouldn't really need anything further implemented after the above, just check it works!
- [x] I'm not sure how much of the manifest is currently implemented in the Python FC code. Check, and open a separate issue for that if necessary - don't consider it part of this one. --> See #148 

Blocked by #95. Part of #23.

MichaelClerx:
Did the `units_of` part, but what next @jonc125 ? Imports are tested in `test_nested_protocols`, as far as I can see

jonc125:
There's the first bit of the description:
> Test by creating a Python version of https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/FunctionalCuration/test/TestS1S2Protocol.hpp

Only nested protocol imports are already tested, not the import usage in `test_S1S2.txt`.

MichaelClerx:
Noble s1s2 test uses unit conversion _within equations_, see discussion on #144

MichaelClerx:
When should `units_of` be resolved, by the way? Is _before_ any model modifications?

jonc125:
Typically it'll be used for variables not given units in the protocol (since otherwise you'd know the units). But if applied to a variable that had specified units I think using those the protocol sets would be most expected?

--
Written on a smartphone; please excuse brevity!
________________________________

MichaelClerx:
We'd better check that then

PR #144, opened 2020-04-01
Started on s1s2 protocol test.

MichaelClerx:
See #106


MichaelClerx:
@jonc125 how much of the C++ code do I need to port, or does test_support do all of that for me?

It seems to run now but with errors. Not sure where to begin!

jonc125:
You can port features iteratively. Basic running of the protocol and checking outputs have expected values will be the same as other Python tests. I'm not sure if we have one that overrides an input, but that can use `Protocol.set_input()`. The manifest will definitely be different because the Python code writes a .h5 file rather than lots of .csv; I can't recall what its manifest generation is like off hand!

MichaelClerx:
Hi @jonc125 , need some help with this!

In particular:
- is it OK to gather up all the units from imported protocols and nested simulations into a single namespace the way we are doing now? (And how do nested simulations define a new model interface or units anyway??)
- what counts as 1 or 2d in the static test files? It doesn't seem to matter what I fill in in `test_s1_s2_lr191`, but I can't get `test_s1_s2` to pass with any combination of 1s and 2s 
- do we need the manifest checking stuff from C++ ? Why was that particular to this S1S2 protocol test and not some general test?
- why does the C++ code appear to change the S1S2 protocol, and do we need to recreate that in these tests? (And if so, how come it passes at the moment?)
- The `check_results` method seems to check the entire results file. Does that mean I won't need the numbers etc. used in the C++ code output checking?

jonc125:
> Is it OK to gather up all the units from imported protocols and nested simulations into a single namespace the way we are doing now? (And how do nested simulations define a new model interface or units anyway??)

Yes, this is fine. In nested simulations these come from nested protocols, and the intended behaviour here is that all the model interface sections get merged into one.

> Do we need the manifest checking stuff from C++ ? Why was that particular to this S1S2 protocol test and not some general test?

It would probably be better in a test specific to the manifest. The S1-S2 test was I think the first whole protocol test written for the C++ implementation, so checking the manifest made sense in that context.

> Why does the C++ code appear to change the S1S2 protocol, and do we need to recreate that in these tests? (And if so, how come it passes at the moment?)

What do you mean by change? It changes a protocol input, which is intentional, and reduces the run time so it works as a unit test (rather than the default longer run in a nightly test).

> What counts as 1 or 2d in the static test files? It doesn't seem to matter what I fill in in `test_s1_s2_lr191`, but I can't get `test_s1_s2` to pass with any combination of 1s and 2s

The reference data for this protocol is all 1d - vectors rather than matrices. You can tell that from the first non-comment line, which specifies num_dims, shape. The 2d load is special in that it assumes 2d with no header line.

> The `check_results` method seems to check the entire results file. Does that mean I won't need the numbers etc. used in the C++ code output checking?

The reference data is for the full S1-S2 protocol. The numbers in the C++ code are for simpler results from the cut-down run. So actually you shouldn't need `test_support.check_results` for this: just check `actual = proto.output_env.look_up(name)` against the hardcoded numbers directly.

fc/protocol.py
```diff
@@ -302,8 +282,13 @@ def process_ns_map(ns_map):
                 self.outputs.extend(imported_proto.outputs)
                 self.plots.extend(imported_proto.plots)
 
-                # Process interface
-                # process_interface(imported_proto.model_interface)
+                # Merge model interface
```
jonc125, line 285: This implementation looks OK to me!

MichaelClerx:
@jonc125 It works for the LR1991 model now, and the full S1S2 protocol on Courtemanche works too.

The Noble model that requires time unit conversion fails the assertions within the protocol*. Could you have a look and see if you know why, or give me some hint on how to start debugging? 

One thing I was wondering about is _when_ the input gets used. As it happens the test code has a set_model _before_ the set_input line, so when we pass in the input the model should already have a time variable in ms, so should be ok?

(_*should we maybe update the error gathering code so that no errors occurring _after_ a protocol assertion are displayed to the user? Assertions are usually sanity checks, no? So that anything that happens after failing one is undefined and most likely irrelevant for debugging?_)

MichaelClerx:
> Can't get the earm noble model to pass in graphstate either...

...but `difrancesco_noble_model_1985` passes graphstate-with-units-conversion, and gives the same errors when used here. So that wasn't it!

MichaelClerx:
OK, _without_ setting the input we get the same errors, so not related to that

MichaelClerx:
I'm finding the error message very hard to read, but as far as I can work out it's this line:
```
E     /home/michael/dev/weblab/fc/fc/library/CardiacLibrary.txt:70:5	    assert fromBasicLib:MultiFold(@2:MathML:min, dVdt) < dvdtThreshold && fromBasicLib:MultiFold(@2:MathML:max, dVdt) > dvdtThreshold
```
so... no steep enough upstroke is detected?

The way the protocol is written it doesn't actually require any conversion on the stimulus current variable, so I'm guessing that's OK.

Voltage is in mV in both model and protocol, so that should be OK too, so does look to be down to time then...

MichaelClerx:
Time units are set using `output oxmeta:time units ms` in the protocol, which seems slightly dubious to me, as we're changing time throughout the protocol so it must also act as an input.
However, adding `independent var units ms` doesn't solve the issue

MichaelClerx:
The model equation uses `oxmeta:membrane_stimulus_current_duration` without defining it as an input or output. How is that not an error? It needs its units converted anyway, I'm guessing

MichaelClerx:
Right. I can make it pass now, but only by modifying the protocol with two extra lines:

```
    independent var units ms
    input oxmeta:membrane_stimulus_current_duration units ms = 2
```

The first makes `fc` use the "input" rule for converting the time variable, which is vital, because we set the time variable when we define the simulation times.

The second makes sure that the `_duration` is in the correct units, although it also overwrites the value which is not something we want to do here.
We can try using `output oxmeta:membrane_stimulus_current_duration units ms` here, but that unit conversion only gets applied _after_ model modifications, so the equation for the stimulus current we define still uses the unconverted variable.

I'm sure I've talked about this before but I don't think we should tie specifying units to the `input` or `output` statements. In the above `time` was clearly used as an output, but it was implicitly an input and requiring the user to think about this (and having a special construct to change the units of the time variable) seems a very silly and unhelpful thing to do.
Secondly, variables used in equations are neither outputs or inputs, but do require unit conversion if we're to have any idea about whether the equations make sense or not. The user also shouldn't have to know about the order of execution within FC protocol interpretation, or about the fact that there's a different conversion rule for inputs and outputs.

jonc125:
An alternative hack might be
```
input oxmeta:membrane_stimulus_current_duration units ms
```
But really we don't want to have to change existing protocols on the Web Lab in order to switch back-ends. At least at first.

So, as partly discussed on Slack, can we:
1. Handle time as an output specially. Perhaps make `_convert_time_if_needed` check `self.outputs` for `oxmeta:time` in a new `else` clause?
2. Try to convert units within equations added by `define`. The challenge here is that we really need the conversions to happen *within* equations, which isn't a use case we'd considered thus far with cellmlmanip, because you have things like `calculation_involving_oxmeta:time <= oxmeta:membrane_stimulus_current_duration` buried inside an `if` clause. So either
    a. we need more complex units conversion code, or
    b. we need to keep variables in defines in their pre-conversion units until after the equations have been changed.

(2a) may be very difficult to code; I'm not sure.

(2b) could have unintended consequences, e.g. causing the same issue in reverse for cases where some variables in equations are created by the protocol.

jonc125:
> One thing I was wondering about is _when_ the input gets used. As it happens the test code has a set_model _before_ the set_input line, so when we pass in the input the model should already have a time variable in ms, so should be ok?

Yes, protocol inputs are only set after the model has been associated with the protocol and the interface changes applied.

> (_*should we maybe update the error gathering code so that no errors occurring _after_ a protocol assertion are displayed to the user? Assertions are usually sanity checks, no? So that anything that happens after failing one is undefined and most likely irrelevant for debugging?_)

That's an interesting idea. We could certainly put an obvious break in the output after the first assertion stating that later errors are likely to be caused by errors above. (We could even put a similar disclaimer at the top of the error output!)

protocols/S1S2.txt
```diff
@@ -54,6 +54,9 @@ units {
 }
 
 model interface {
+    independent var units ms
```
jonc125, line 57: As noted in the comments, I'd rather not have to change protocols to get them working.

MichaelClerx:
>     1. 
> 
>     2. Try to convert units within equations added by `define`. The challenge here is that we really need the conversions to happen _within_ equations, which isn't a use case we'd considered thus far with cellmlmanip, because you have things like `calculation_involving_oxmeta:time <= oxmeta:membrane_stimulus_current_duration` buried inside an `if` clause. So either
>        a. we need more complex units conversion code, or
>        b. we need to keep variables in defines in their pre-conversion units until after the equations have been changed.
> 
> 
> (2a) may be very difficult to code; I'm not sure. This is what pycml does, however, so is guaranteed to retain the same behaviour.
> 
> (2b) could have unintended consequences, e.g. causing the same issue in reverse for cases where some variables in equations are created by the protocol.

I'll get started on (1), but I don't like (2a) or (2b). As a user, and keeping in mind that the WL is meant to make modelling work more trustworthy, I want predictability and control.
So some very clever strategy that guesses the units we mean instead of us specifying them directly seems the to wrong way to go?
I'd be much happier with a WL where the current S1S2 protocol gives an error when trying to evaluate the `define`

jonc125:
Personally I think the current pycml *is* predictable and reliable in its behaviour here. Variables are either in the units the model gives them, or in the units the protocol specifies if it overrides them, and conversions are applied wherever needed. The protocol author can have as much control as they like. It's just more difficult for an implementation...

MichaelClerx:
Can I amend that to "predictable and obvious"?

> and conversions are applied wherever needed

This bit is it tricky! The user would have to have some idea of the algorithm we use. I think that only a user who is quite involved would find it predictable

MichaelClerx:
I mean, technically speaking the latex compiler is predictable. That's not what I mean though. I want it to be really easy and obvious

jonc125:
I don't think you need to know the algorithm, just that it ensures correctness? So any operation involves operands in the same units, for instance.

MichaelClerx:
> I don't think you need to know the algorithm, just that it ensures correctness? So any operation involves operands in the same units, for instance.

Yeah but how does that propagate through a big equation? Is there only one unique way to resolve each equation? Will multiple conversion factors always cancel out, or can we get numerical leftovers? What about log(1 V / 2 kV), does that get a conversion factor in it?

MichaelClerx:
Another thing is that CellML emphatically does _not_ convert within equations, so users that are writing their own models will need to remember that the rules for CellML and the rules for updating CellML from FC (with MathML:log and everything) are not the same

jonc125:
I guess there might be numerical leftovers, but then there might if you specify the units for each variable so no difference there I think. Your log example would need a conversion factor for the argument.

The difference from CellML (2) is a weightier argument. Although CellML 1 allowed it and pycml has always done it, so there is still a change in behaviour.

What would you want to happen if an author uses a variable in an equation *without* specifying units for it?

MichaelClerx:
I'm thinking I'd want FC to complain if you used variables in an equation but didn't specify what units they were meant to be in.
Although I suppose there's a case for not specifying _any_ units e.g. when writing a stimulus current in terms from the model (but then what happens if some of those terms are missing, and you want to give default values, how will FC know what units _they_ should be in?)

Actually, how does that work for plotting with time specifically? If I just say run for 1000 time units on a model in second and one in ms, what does the graph look like?

jonc125:
One issue with changing the behaviour is that `SinglePace.txt` has this stimulus definition, and it's used by lots of other protocols, so we'd need to change about half the standard protocols currently on Web Lab, and probably quite a few users have created based on them.

There are also things like
```
define oxmeta:membrane_rapid_delayed_rectifier_potassium_current_conductance = local:conductance_scale_factor * original_definition
```
What should happen here?

With optional variables there's always another source for the units; either an input/output statement, or the units of the default clause itself.

With plots if you want consistent comparisons between models you need to specify units. That's why we let users do it :)

MichaelClerx:
What would you suggest instead?

We could handle it in the sympy conversion I suppose?
Change the RHS of any binary operator to make sure its units match the LHS?

jonc125:
What pycml does for FC is convert within equations only for those added or altered by protocols, since as you say CellML models don't require it so model authors probably haven't allowed for it.

Binary operators don't have an LHS & RHS so you have to be a little more clever. Pycml has strategies for each operator (the `_set_in_units` methods), and actually propagates sets of 'possible units' across the whole expression tree so that it can avoid the cancelling conversion factors issue as much as possible, by minimising the number of conversions needed to maintain consistency at each operation. But you can just convert operands for each operator, working down the expression tree, with strategies such as:
- for relational operations, set all operands to have the same units, choosing those with the smallest multiplicative factor when expanded (or for top-level `eq`, convert from RHS units to LHS units). The resulting expression has boolean 'units' (or dimensionless ;) )
- for times & divide, the parent expression dictates the desired units, and the result is converted to those
- for plus & minus, all operands are converted to the units desired by the context
- for trig, exp, log, abs, rem, set the operand(s) to have dimensionless units
- for piecewise, the conditions have to be boolean, and the pieces all have the units desired for the overall expression
- for power & root, the exponent/degree is dimensionless, and the result is converted to the desired units

MichaelClerx:
That's a lot of added complexity for a very small pay off. How about if someone write V + mV we just say "you made an error, please fix"

jonc125:
Because it means altering half the protocols already on WL :)

jonc125:
And I think we need to do it to make the `original_definition` construct work, anyway.

jonc125:
If you want to focus on other issues for now, we can merge this PR without the units conversion case, and I can try to take a look at this over the next week?

MichaelClerx:
If you can't use the `original_definition` construct without getting inconsistent units then you shouldn't.

And I don't mind changing the protocols if it makes things clearer.

The point is to make the things that people do when modelling more reliable. That means making easy-to-use tools that users can easily understand. Things should just work the way people without WL experience expect them to at first glance (insofar as that is possible). The more complicated stuff we do under the hood, the more room we create for tricky intricacies and gotchas and bugs etc. That's fun if you like programming, but doesn't fit the brief.

MichaelClerx:
> If you want to focus on other issues for now, we can merge this PR without the units conversion case, and I can try to take a look at this over the next week?

OK I'll get the time units conversion in and then leave it.

protocols/S1S2.txt
MichaelClerx: Changes undone!

MichaelClerx:
OK, ready for re-review then!

Issue #99, opened 2020-02-17
Support test_INa_IV_curves protocol


jonc125:
Part of #23, following from #95 and #96 and #97 .

See https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/FunctionalCuration/test/TestOptionalVariables.hpp for tests of this protocol in the C++ code. We'll need to create equivalent testing in Python.

- [x] Implement `DeclareVariable` (for `var ... units ... = ...`) #154
- [x] Implement units conversion rules (`convert ... to ... by ...`; `UnitsConversion` action class) #156 

Up to here will allow testing on models that have `oxmeta:extracellular_sodium_concentration`, `oxmeta:cytosolic_sodium_concentration`, `oxmeta:cytosolic_potassium_concentration`, and `oxmeta:cytosolic_calcium_concentration`. Make sure current units conversions are needed!

Once these steps are done, `test_ical_proto.py` should also work. https://chaste.cs.ox.ac.uk/trac/browser/chaste/projects/FunctionalCuration/test/TestIcalProtocol.hpp gives ideas for a shorter version of this test.

- [x] Implement `OptionalVariable` handling (for `optional ...`) for clamped inputs. #157 

- [x] Once all the above is done, we may find that `test_clamping.py:test_clamp_to_data_file` just works. If not, open a separate issue for fixing what parts of it still need implementing.

MichaelClerx:
Do we have a strategy for generating names for local (``var``) variables in error messages? Usually we'd use rdf terms but they don't have to have any.

jonc125:
I'd guess report just the variable name? Since that's the name given in the protocol so will be what the user expects?

MichaelClerx:
> I'd guess report just the variable name? Since that's the name given in the protocol so will be what the user expects?

Well, you might not remember what the local name for it was at the point that you're creating the message, because its name in the model will be e.g. `protocol__x_3` instead of `x`

MichaelClerx:
@jonc125 what do I do with ``var`` variables from nested/imported protocols? 

I'm guessing we need some sort of mechanism so that each protocol has its own namespacing / scope for these?

jonc125:
[...]

PR #153, opened 2020-04-08
Run INa IV protocols

MichaelClerx:
See #99 

Still needs

- [x] unit conversion rules #156 
- [x] clamping optional values to initial variables #177

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -1914,14 +1914,14 @@ class ModelInterface(BaseGroupAction):
                         raise ProtocolError(f'Clamp-to-initial-value set for unknown variable {pvar.long_name}.')
 
                 # Evaluate and update equation
-                defn = self.model.get_definition(var)
-                if defn.lhs.is_Derivative:
+                if self.model.is_state(var):
                     value = var.initial_value
                 else:
-                    value = defn.rhs.evalf()  # TODO: What if there are variable references in the RHS?
+                    value = self.model.get_value(var)
                 new_defn = sympy.Eq(var, self.model.create_quantity(value, var.units))
-                if defn is not None:
-                    self.model.remove_equation(defn)
+                old_defn = self.model.get_definition(var)
+                if old_defn is not None:
+                    self.model.remove_equation(old_defn)
                 self.model.add_equation(new_defn)
 
     def _process_transitive_variables(self):
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -1914,10 +1914,7 @@ class ModelInterface(BaseGroupAction):
                         raise ProtocolError(f'Clamp-to-initial-value set for unknown variable {pvar.long_name}.')
 
                 # Evaluate and update equation
-                if self.model.is_state(var):
-                    value = var.initial_value
-                else:
-                    value = self.model.get_value(var)
+                value = self.model.get_value(var)
                 new_defn = sympy.Eq(var, self.model.create_quantity(value, var.units))
                 old_defn = self.model.get_definition(var)
                 if old_defn is not None:
--- a/test/test_iv_curves.py
+++ b/test/test_iv_curves.py
@@ -7,7 +7,6 @@ import pytest
 import fc
 
 
-@pytest.mark.xfail(strict=True, reason='No pycml replacement yet')
 def test_clamping_with_old_model():
 
     proto = fc.Protocol('test/protocols/test_INa_IV_curves.txt')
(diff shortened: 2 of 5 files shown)


MichaelClerx:
Why is this called "optional variables"? Doesn't include any optional inputs/output, but instead tests `var` declarations and unit conversion rules.

jonc125:
One test protocol has things like `optional oxmeta:extracellular_sodium_concentration`, and another things like
```
optional oxmeta:temperature default 310 :: K
optional oxmeta:sodium_reversal_potential default (R*oxmeta:temperature/F) * MathML:ln(oxmeta:extracellular_sodium_concentration / oxmeta:cytosolic_sodium_concentration)
```

It's just those aren't the only new features in those protocols. You could create new smaller test cases for individual features if you like.

MichaelClerx:
@jonc125 I think this failure has nothing to do with this PR but must come from here:

```
    def __init__(self):
        """Initialise the parser."""
        # We just store the original stack limit here, so we can increase
        # it for the lifetime of this object if needed for parsing, on the
        # basis that if one expression needs to, several are likely to.
        self._stack_depth_factor = 1
        self._original_stack_limit = sys.getrecursionlimit()
        self.increase_stack_depth_limit()

    def __del__(self):
        """Reset the stack limit if it changed."""
        sys.setrecursionlimit(self._original_stack_limit)

    def increase_stack_depth_limit(self, step=0.5):
        """Increase the limit by the given factor of the original."""
        self._stack_depth_factor += 0.5
        new_limit = int(
            self._stack_depth_factor * self._original_stack_limit)
        print('Increasing recursion limit to', new_limit,
              file=sys.stderr)
        sys.setrecursionlimit(new_limit)
```

I guess every CSP created in the same process increases the stack limit further? Seems a bad idea...

MichaelClerx:
Got rid of the CSP thing, but 2/3 tests still fail. No idea why or what those error messages mean.
Can you take this one up, @jonc125 ?

MichaelClerx:
Remaining issue seems to be https://github.com/ModellingWebLab/cellmlmanip/issues/278

fc/parsing/actions.py
```diff
@@ -1914,14 +1914,14 @@ def _handle_clamping(self):
                         raise ProtocolError(f'Clamp-to-initial-value set for unknown variable {pvar.long_name}.')
 
                 # Evaluate and update equation
-                defn = self.model.get_definition(var)
-                if defn.lhs.is_Derivative:
+                if self.model.is_state(var):
                     value = var.initial_value
                 else:
-                    value = defn.rhs.evalf()  # TODO: What if there are variable references in the RHS?
+                    value = self.model.get_value(var)
```
jonc125, line 1920: We could just use `value = self.model.get_value(var)` without the check now!

test/test_iv_curves.py
```diff
@@ -0,0 +1,60 @@
...
+    proto.run()
+
+    # Check the key outputs haven't changed.
+    peaks = proto.output_env.look_up('normalised_peak_currents').array
+    expected = [1.0, 0.999998, 0.999981, 0.999906, 0.999564, 0.999062, 0.99795, 0.995344, 0.988466, 0.966859, 0.887918,
+                0.624599, 0.192671, 0.0186169, 0.00193235, 0.00116954, 0.00114086, 0.00113976, 0.00113976]
+    assert len(peaks) == len(expected)
+    ipeaks = iter(peaks)
+    for value in expected:
+        assert pytest.approx(value, 1e-3) == next(ipeaks)
```
jonc125, line 24:
```suggestion
        assert next(ipeaks) == pytest.approx(value, 1e-3)
```

test/test_iv_curves.py
```diff
@@ -0,0 +1,60 @@
...
```
jonc125, line 42:
```suggestion
        assert next(ipeaks) == pytest.approx(value, 1e-3)
```

test/test_iv_curves.py
```diff
@@ -0,0 +1,60 @@
...
+    proto.run()
+
+    # Check the key outputs haven't changed.
+    peaks = proto.output_env.look_up('normalised_peak_currents').array
+    expected = [0.893527, 0.946238, 0.980761, 0.999288, 1, 0.98876, 0.961362, 0.903678, 0.79484, 0.622032, 0.410937,
+                0.224468, 0.104744, 0.044427, 0.018137, 0.00741457, 0.00312374, 0.000728363, 0.000343776]
+    assert len(peaks) == len(expected)
+    ipeaks = iter(peaks)
+    for value in expected:
+        assert pytest.approx(value, 1e-3) == next(ipeaks)
```
jonc125, line 59:
```suggestion
        assert next(ipeaks) == pytest.approx(value, 1e-3)
```

Issue #154, opened 2020-04-08
Add tests and support for `var` declarations


MichaelClerx:
- var with initial value (constant)
- var with initial value, with ODE set by  `define`
- var without initial value, with constant set in `define`
- var without either --> error
- var with two defines --> error

PR #155, opened 2020-04-08
Add support for var declarations.

MichaelClerx:
See #154


fc/parsing/actions.py
```diff
@@ -904,6 +912,11 @@ class ModelInterface(BaseGroupAction):
         A list of :class:`ClampVariable` instances. These are created for e.g. ``clamp x``, while statements with an RHS
         such as ``clamp x to 1`` are treated as an alias of ``define x = 1`` and do not lead to creation of a
         ``ClampVariable`` object.
+    ``local_var_declarations``
+        A list of :class:`DeclareVariable` instances. The are created for ``var x`` statements.
```
jonc125, line 916: The -> They ?

Issue #141, opened 2020-03-31
Make test_other_ontologies pass


MichaelClerx:
> test_other_ontologies has terms not in `oxmeta` and a generic vector output (not the state vector) - should be covered by #95 and #97 actually

Tried it, but doesn't work out of the box, so made this ticket for it

jonc125:
`test_annotating_with_other_ontologies` in `test_various_protocols.py` would also come under this topic.

MichaelClerx:
> `test_annotating_with_other_ontologies` in `test_various_protocols.py` would also come under this topic.

Same test, I think?

MichaelClerx:
Might need splitting, that test, as it also tests annotating outputs with categories?

MichaelClerx:
Or rename something like 'test_advanced_annotation' :D

jonc125:
Fair enough :)

MichaelClerx:
Closed in #174

Issue #157, opened 2020-04-09
Optional variables in model interface: add tests and support


MichaelClerx:
Required for #99

- Optional input: don't create if optional
- Optional output: not sure if we need to do much

`INa_IV_curves` contains examples of optional inputs, an optional output (with a default value), and an optional variable that isn't an input or output

MichaelClerx:
Bit confused about the rules on https://chaste.cs.ox.ac.uk/trac/wiki/FunctionalCuration/ProtocolSyntax#Modelinterface

About inputs:

> If the variable doesn't exist, this is an error unless units are given - the variable is then created.

> Unless the variable is specified as optional (see below) it must also be given a value, either using an initial value here, or defining it by a define clause, or both (if defining by an ODE).

In the comments in the code for FC we seem to interpret this as "optional inputs do not need to be created", but the first sentence here contradicts that.

About outputs:

> Nothing is written about optional outputs!

About optional itself:

> If an optional variable doesn't appear in the model, then it is not an error for it to be referenced by an input, output, clamp or define construct. If a default clause is included, then it will be used as the variable's definition; otherwise any construct referencing the variable will simply be ignored.

This just seems to say "create any optional variable that has a default value", regardless of whether it's an input or output or neither?

MichaelClerx:
But then it says

> Note that this 'optional-ness' only applies directly to uses within this section of the protocol - if a simulation or post-processing tries to access or set an optional variable that doesn't exist, this is still an error. The exceptions are post-processing assignments and output specifications that are themselves guarded by an optional keyword. 

Which would make optional outputs pointless, because even though they get created, you're not allowed read them out later?

MichaelClerx:
Protocols such as `INa_IV_curves` _do_ refer to optional outputs (with default values) in their post-processing though

MichaelClerx:
Not sure how to create the optional variables now:

```
    var R units mJ_per_Mol_K = 8314.4621
    var F units C_per_Mol = 96485.3365
    optional oxmeta:temperature default 310 :: K
    optional oxmeta:sodium_reversal_potential default (R*oxmeta:temperature/F) * MathML:ln(oxmeta:extracellular_sodium_concentration / oxmeta:cytosolic_sodium_concentration)
    output oxmeta:sodium_reversal_potential units mV
```

To create a new variable for `oxmeta:temperature`, I need to know its units
The only way to get this seems to be to inspect the RHS.
[...]

PR #158, opened 2020-04-09
Parsing optional statements in model interface.

### LINKED ISSUE #105: Consistency checks for model interface
Have some sanity checks before/after modification that report errors helpfully for the user.

- [x] ~Double~ Inconsistent inputs
- [x] ~Double~ Inconsistent outputs
- [x] Input/output overlap: must have same units
- [x] Check that a variable isn't defined more than once in the interface. Have a check before making changes that we don't have 2 define and/or clamp clauses targeting the same variable, and produce a friendly error.

MichaelClerx:
Addresses #157. Fixes #105.

Significant rewrite of ModelInterface.modify_model(), based on https://github.com/ModellingWebLab/weblab-fc/issues/157#issuecomment-614859533

- Before modifications begin, information from inputs, outputs, defines, clamp-tos, and var statements is collated and stored in ProtocolVariable objects.
- Where possible, ProtocolVariables are resolved to model variables, and sanity-checking is performed
  - To allow nesting, this is done slightly more leniently than before, e.g. multiple `input` declarations are allowed on the same variable, as long as they don't make conflicting demands
- Unit conversion of inputs is slightly rewritten, but more or less the same: I wanted to do outputs here too but this caused more bookkeeping due to transitive variables & the `state_variable` annotation
- Variable creation requires units, units must sometimes be inferred from the RHS, and this requires the RHS in sympy which requires all RHS variables to exist. To deal with this situation, creating and modifying variables is now performed in a single method, and in several passes. The system is deemed irresolvable if a single pass is unable to make any progress.


MichaelClerx:
The s1_s2_noble test causes an error in cellmlmanip, where the `to_unit` becomes `None` but a conversion is still triggered, seems to go wrong here:

https://github.com/ModellingWebLab/cellmlmanip/blob/877918347d9d9e7be08692ba5bcc01780175eb71/cellmlmanip/units.py#L775-L778

```python
            if to_units is not None or was_converted:
                # Convert the result if required
                expr = expr.func(*new_args)
                expr, was_converted, actual_units = maybe_convert_expr(expr, was_converted, actual_units, to_units)
```

These lines do look wrong to me: maybe_convert_expr() can be called with `to_units=None` which seems to cause this error?

MichaelClerx:
Seems to be a separate issue, as I get the same error locally with the master branch of `fc`

MichaelClerx:
Made a separate branch to reproduce: https://github.com/ModellingWebLab/weblab-fc/pull/166

MichaelClerx:
Stopped this from happening now, but the lines above still look dubious to me @jonc125 :D

MichaelClerx:
I think the bit with the output variables can be tidied up a bit further:

- make `get_variables_transitively` return variables + terms, apply this to outputs in `collate` to create ProtocolVariables for these outputs
- stop calling `get_variables_transitively` in `_convert_output_units` (now becomes a simple loop)
- store outputs better, and use this in protocol.py to pass to code generation, stop calling `get_variables_transitively` in `code_generation.py`
- presumably the rule is "If a non-optional output doesn't resolve directly, but does resolve transitively, then that's fine."?

MichaelClerx:
Wondering if `ModelInterface._expr` and `ModelInterface.merge` can be simplified, and we no longer need to store all the base actions for inputs/outputs/optional etc.? (At least not separately)
Also duplicates would be fine in current code, so merge can be easier?

fc/parsing/actions.py
```diff
@@ -882,6 +885,134 @@ def _expr(self):
...
+            self.aliases.append(pvar.name)
+
+        # Merge type information
+        self.input = self.input or pvar.input
+        self.output = self.output or pvar.output
+        self.optional = self.optional or pvar.optional
+        self.local = self.local or pvar.local
+
+        # Check type information isn't conflicting
+        assert not self.local and (self.input or self.output), f'{self.long_name} is local AND input/output'
```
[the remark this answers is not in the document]
MichaelClerx: I think there's no way for the user to create this situation at the moment.

fc/parsing/actions.py
```diff
@@ -1200,204 +1266,301 @@ def _number_generator(self, value, units):
...
-                var = lhs.args[0]
-
-                # Initial value must be set via input or var, or already be set (if this was already a state variable)
-                if var.initial_value is None and var not in initial_values:
-                    terms = '/'.join(str(x) for x in self.model.get_ontology_terms_by_variable(var))
-                    raise ProtocolError(f'Variable {terms} is being set as a state variable but has no initial value'
-                                        ' (this can be set using an `input` or `var` statement)')
-            else:
-                var = lhs
+        # Check if the units are set by an input or output instead or in addition to the independent var clause
```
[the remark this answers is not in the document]
MichaelClerx: I guess we could start `_collate_variable_information` by checking any/all `independent var units` clauses and adding them as inputs?

fc/parsing/actions.py
```diff
@@ -1200,204 +1266,301 @@ def _number_generator(self, value, units):
...
+            pvar.merge_initial_value(ref.initial_value)
 
-                if variable is self.time_variable and output.units is not None:
-                    units = self.units.get_unit(output.units)
-                    self.time_variable = self.model.convert_variable(variable, units, DataDirectionFlow.INPUT)
-                    break
+        # Add outputs
+        #   output <prefix:term> [units <uname>]
+        for ref in self.outputs:
+            if ref.prefixed_name == 'oxmeta:state_variable':    # Skip 'magic' oxmeta:state_variable term
```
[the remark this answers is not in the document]
MichaelClerx: Fixed now! But it _is_ still special in other ways

fc/parsing/actions.py
```diff
@@ -1451,7 +1616,7 @@ def _list_parameters(self):
 
             # Parameters are inputs that aren't states, and have a constant RHS
             eq = self.model.get_definition(variable)
-            if isinstance(eq.lhs, VariableDummy) and len(eq.rhs.atoms(VariableDummy)) == 0:
+            if not eq.lhs.is_Derivative and len(eq.rhs.atoms(VariableDummy)) == 0:
```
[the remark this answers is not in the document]
MichaelClerx: Just a general sense that checking a property is better than using type checking?
Been a while since I saw this code so learned a bit more about cellmlmanip / sympy

fc/parsing/actions.py
```diff
@@ -1442,6 +1605,8 @@ def _convert_output_units(self):
     def _list_parameters(self):
         """Populates the list of model parameters (inputs with a constant RHS)."""
 
+        # TODO: Is it OK if this list includes multiple ontology terms mapped to the same variable? If not, then
```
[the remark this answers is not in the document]
MichaelClerx: Ah. Only seeing this now after changing it. (The relevant code is now in `code_generation.py`).
Currently trying to go from the ProtocolVariable objects for everything, so I suppose that could store all alias rdf_terms it knows and add them all as possible inputs?

fc/parsing/actions.py
```diff
@@ -1200,204 +1266,301 @@ def _number_generator(self, value, units):
...
+                                            ' initial value (this can be set using an `input` or `var` statement).')
+
+                    # Remove existing equation
+                    if eq is not None:
+                        self.model.remove_equation(eq)
+
+                    # Add or replace equation
+                    # If required, convert units within RHS to make it consistent and match the LHS units
+                    eq = sympy.Eq(lhs, rhs)
+                    eq = self.units.convert_expression_recursively(eq, None)
```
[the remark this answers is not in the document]
MichaelClerx: Updated this now, both seem to do the same in our unit tests!

fc/parsing/actions.py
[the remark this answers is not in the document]
MichaelClerx: They would have to create an input/output/optional that's not in a namespace, or a local variable that is?
Alternatively a local variable called e.g. `x:y`, but not sure the parser allows that?

fc/parsing/actions.py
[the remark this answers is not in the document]
MichaelClerx: But I'm also happy handling this particular statement separately. It makes for better error messages anyway. And the `independent var units` statement works without annotations so not part of the input/output/optional/local roundup?

fc/parsing/actions.py
```diff
@@ -1200,204 +1266,301 @@ def _number_generator(self, value, units):
...
+                except MissingVariableError:
+                    raise ProtocolError(f'Define or clamp statement found for unknown variable: {name}.')
 
-    def _add_input_variables(self):
-        """Ensure requested input variables exist, unless they are optional.
+                # Create new protocol variable to store info
+                pvar = self.protocol_variables[name] = ProtocolVariable(name)
 
-        If the variable exists its units are checked and converted if needed.
+            # Any variable that is redefined by an equation is treated as an input
```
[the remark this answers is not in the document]
MichaelClerx: My reasoning was that if you redefine a variable, (and you suggested on slack that this could be done for inputs, outputs, or anything in the protocol), then it needs to be units-converted using the `input` rather than `output` rule.
Also, we're changing its value, which to me makes it an input.
(Similarly I'd argue time is an input)

MichaelClerx:
I tried to fix #141 and make handling of "transitive outputs" a bit more regular. Turns out I didn't actually fix 141 in the end, but I think (hope) that's actually due to other ontologies being used instead of the transitive outputs.

Bigger changes since last review:

- code generation now works off of ProtocolVariables instead of input/output lists (this makes sense if we do plan to deprecate these constructs)
- ProtocolVariable objects now can have a list of "transitive variables", and have the "output_category" property
- The various merge() methods on ProtocolVariable are now a single update() method
- protocol_variables is a list, not a dict
- If a "magic" ProtocolVariable exists that keeps track of the state variables, this is known to the ModelInterface.
- output unit conversion is no longer a separate thing
- ordering of state variables is rewritten. As before, it attempts to retain the original order of state variables, pre-modification. I think this version is slightly different (but better?) than the last though (as far as I can work out in my head): The previous version estimated an order after input variables were converted, but before output variables were converted. This would mean that an output V in a different unit was still variable 0, but an input V in a different unit might change order? I _think_ the old version did that, apologies if it didn't, the new version _should_ maintain the order no matter what, with any new state variables added to the end of the list

fc/parsing/actions.py
```diff
@@ -1200,204 +1278,353 @@ def _number_generator(self, value, units):
...
-            # missing variable and LHS are optional. In which case, just skip the equation (but log it).
-            eqs.append(eq.to_sympy(self._variable_generator, self._number_generator))
-        self.sympy_equations = eqs
+    def _annotate_state_variables(self):
+        """
+        Annotate all state variables with the 'magic' oxmeta:state_variable term.
 
-    #######################################
-    # Helper methods for model manipulation
+        This method also ensures that all state variables have an rdf identity (``var.rdf_identity``).
```
MichaelClerx, line 1285: (This means that we can still identify them after unit conversion!)

fc/code_generation.py
```diff
@@ -195,42 +193,48 @@ def derivative_name(deriv):
...
+        if pvar.output and pvar.model_variable is not None:
+            # Single variable output
+            var_name = variable_name(pvar.model_variable)
+            length = None
+            output_variables.add(pvar.model_variable)
+        elif pvar.output_category and pvar.transitive_variables:
             # Vector output
-            if term in vector_orderings:
-                order = vector_orderings[term]
+            # TODO: Can a variable be both?
```
jonc125, line 224: Yes, it can. Membrane voltage is the most common example!

test/test_code_generation.py
```diff
@@ -54,19 +54,27 @@ def test_generate_weblab_model(tmp_path):
     model = os.path.join('test', 'models', 'hodgkin_huxley_squid_axon_model_1952_modified.cellml')
     model = cellmlmanip.load_model(model)
 
-    # Select model outputs
-    outputs = [
-        VariableReference.create('oxmeta', OXMETA_NS, 'membrane_fast_sodium_current'),
```
jonc125, line 59: I've not reviewed actions.py yet. Do we still need the `create` method if this code is gone?

test/test_code_generation.py
```diff
@@ -54,19 +54,27 @@ def test_generate_weblab_model(tmp_path):
...
+    variables = [
+        (True, False, 'membrane_fast_sodium_current_conductance'),
+        (True, False, 'membrane_potassium_current_conductance'),
+        (False, True, 'membrane_fast_sodium_current'),
+        (False, True, 'membrane_voltage'),
+        (False, True, 'time'),
     ]
+    for input, output, name in variables:
+        rdf_term = create_rdf_node((OXMETA_NS, name))
+        pvar = ProtocolVariable(OXMETA_NS + ':' + name, name, rdf_term)
```
jonc125, line 68: Shouldn't the first argument be `'oxmeta:' + name`?

test/test_code_generation.py
```diff
@@ -54,19 +54,27 @@ def test_generate_weblab_model(tmp_path):
...
+        pvar.update(input=input, output=output)
+        pvar.update(model_variable=model.get_variable_by_ontology_term(rdf_term))
+        protocol_variables.append(pvar)
+
+    # State variable output
+    rdf_term = create_rdf_node((OXMETA_NS, 'state_variable'))
+    pvar = ProtocolVariable(OXMETA_NS + ':' + 'state_variable', 'state_variable', rdf_term)
```
jonc125, line 75: Ditto here?

fc/code_generation.py
jonc125: Unless you mean can an RDF term be both, in which case the answer is no.

MichaelClerx: I think in actions.py I've left the opportunity occasionally for a term to "is" a variable, while it also has transitive variables. Is that something that can go then?

fc/code_generation.py
```diff
@@ -195,42 +193,48 @@ def derivative_name(deriv):
...
-        variable = model.get_variable_by_ontology_term(parameter.rdf_term)
-        parameter_info.append({
-            'index': i,
-            'local_name': parameter.local_name,
-            'var_name': variable_name(variable),
-            'initial_value': model.get_value(variable),
-        })
-        parameter_variables[variable] = i
+    for pvar in protocol_variables:
+        if pvar.input and pvar.model_variable is not None:
```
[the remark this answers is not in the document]
MichaelClerx: Would like that but it depends on the `model`. So could still be an `is_parameter(model)` method, I guess?

fc/parsing/actions.py
```diff
@@ -882,6 +878,150 @@ def _expr(self):
...
+        if model_variable is not None:
+            if self.model_variable is None:
+                self.model_variable = model_variable
+            else:
+                assert model_variable == self.model_variable, 'ProtocolVariable representing multiple model variables'
+
+        # Add transitive variables
+        if transitive_variables:
+            self.transitive_variables.update(transitive_variables)
+
```
jonc125, line 1003: We should probably `assert not self.transitive_variables or self.model_variable is None` here since it shouldn't be possible for one reference to be both a single variable and a vector.

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
-    def _list_parameters(self):
-        """Populates the list of model parameters (inputs with a constant RHS)."""
-
-        for var in self.inputs:
+        # Gather state variables
+        states = get_variables_that_are_version_of(self.model, STATE_ANNOTATION)
+
+        # Create and store ordering, preserving original ordering as much as possible
+        order = []
+        todo = set([var.rdf_identity for var in states])
```
jonc125, line 1646: Conversion to set loses ordering here, and isn't guaranteed to be consistent across runs either. The former isn't critical, but the latter is!

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
+        for rdf_identity in original_ordering:
             try:
-                variable = self.model.get_variable_by_ontology_term(var.rdf_term)
+                todo.remove(rdf_identity)
             except KeyError:
-                # Skip optional variables
                 continue
+            order.append(rdf_identity)
+        order.extend(todo)
```
jonc125, line 1653: So we'd need to ensure we sort `todo` into a predictable order here; perhaps just a string sort on the contents? So `order.extend(sorted(todo))`?

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
+        self.vector_orderings[STATE_ANNOTATION] = {rdf_identity: i for i, rdf_identity in enumerate(order)}
 
-            # Parameters are inputs that aren't states, and have a constant RHS
-            eq = self.model.get_definition(variable)
-            if isinstance(eq.lhs, VariableDummy) and len(eq.rhs.atoms(VariableDummy)) == 0:
-                self.parameters.append(var)
+        # Set transitive variables for `state_variable` output, if present
```
jonc125, line 1656: I'd add 'in the protocol' to this comment, for clarity.

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
-        """Remove model equations and variables not needed for generating desired outputs.
+    def _purge_unused_mathematics(self):
+        """Remove model equations and variables not needed for generating desired outputs."""
 
-        :param output_variables: the set of variables appearing in outputs, either directly or as part of a vector
-        """
+        # Output variables
+        required_variables = set()
+        for pvar in self.protocol_variables:
+            if pvar.output or pvar.output_category:
```
jonc125, line 1666: It occurs to me that we should probably remove this line, since input variables should be considered required. It'd be a pretty bizarre protocol that defined an input which wasn't used in computing any outputs, but if someone did do so they'd be pretty confused as to why their input didn't exist!

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
+                required_variables.update(pvar.transitive_variables)
+
+        # Symbols used directly in equations computing outputs
         import networkx as nx
         graph = self.model.graph
-        required_variables = set(output_variables)
+        for variable in list(required_variables):
+            required_variables.update(nx.ancestors(graph, variable))
 
         # Time is always needed, even if there are no state variables!
```
jonc125, line 1677: Looking a few lines below, this means we don't need to add it in the loop over derivatives any more. Forgot to remove that when I added this!

fc/parsing/actions.py
```diff
@@ -1413,63 +1630,53 @@ def _handle_clamping(self):
...
-            self.vector_orderings[state_annotation][state_var.rdf_identity] = i
-
-    def _convert_output_units(self):
-        """Convert units for all outputs if needed.
-
-        :return: The set of variables appearing in outputs, either directly or as part of a vector, in the desired
-            units.
+    def _gather_state_variables(self, original_ordering):
+        """
+        Gather all variables annotated as states, and create a vector ordering. If required, update the ProtocolVariable
```
jonc125, line 1635: I like this idea.

fc/parsing/actions.py
```diff
@@ -1052,111 +1197,31 @@ def modify_model(self, model, units):
...
-        # Ensure inputs exist, adding them if needed, and storing initial values set by user.
-        # This also performs unit conversion where needed for existing input variables.
-        self._add_input_variables()
-
-        # Add variables created by ``var`` statements
-        self._add_local_variables()
+        # Annotate all state variables with the magic `oxmeta:state_variable` term. This is done before unit conversion
+        # so that annotations are transferred where needed. The original order in which state variables were defined is
+        # stored.
+        original_order = self._annotate_state_variables()
```
jonc125, line 1203: Maybe call the variable `original_state_order`?

jonc125:
> I tried to fix #141 and make handling of "transitive outputs" a bit more regular. Turns out I didn't actually fix 141 in the end, but I think (hope) that's actually due to other ontologies being used instead of the transitive outputs.

Hmm, I think using other ontologies should work, at least in code generation, so I would be worried by this. Haven't got as far in my review as running the test to see what happens yet though!

fc/parsing/actions.py
```diff
@@ -1200,204 +1268,353 @@ def _number_generator(self, value, units):
...
+                name_to_pvar[ref.prefixed_name] = pvar
 
-        name = self.model.get_unique_name('protocol__' + var.local_name)
-        units = self.units.get_unit(var.units)
-        variable = self.model.add_variable(name, units)
-        self.model.add_cmeta_id(variable)
-        self.model.rdf.add((variable.rdf_identity, PRED_IS, var.rdf_term))
-        return variable
+                # Store 'state_variable' pvar, if used in protocol
+                if ref.rdf_term == STATE_ANNOTATION:
```
jonc125, line 1299: I wasn't convinced by this at first, but I think I'm coming round to the idea!

fc/parsing/actions.py
```diff
@@ -1200,204 +1268,353 @@ def _number_generator(self, value, units):
...
+            pvar.update(input=True, units=ref.units, initial_value=ref.initial_value)
 
-        If it doesn't exist but is marked as optional, nothing is done.
+        # Add outputs and output categories
+        #   output <prefix:term> [units <uname>]
+        for ref in self.outputs:
+            pvar = get(ref)
+            pvar.update(output=True, units=ref.units)
+
+            # Check for "transitively" connected variables (e.g. the rdf term represents a category)
```
jonc125, line 1317: OK, here we definitely have a problem, as the transitive search also (by design) picks up directly annotated variables, so you'll get length 1 vectors appearing here for every output variable.

I'm also very wary of treating the state vector differently from other vector outputs: doing so is likely to hide bugs with the general case. I _suspect_ (but haven't finished reviewing yet) doing the transitive search here for the general case means you're failing to update correctly after unit conversion, and this isn't picked up in the state vector case because it has magic support. In pycml the transitive lookup happens *after* all other manipulations.

fc/parsing/actions.py
```diff
@@ -1200,204 +1268,353 @@ def _number_generator(self, value, units):
...
-        # Get all variables named by inputs and outputs
-        # Note: Don't worry about clashes here, this is checked later in self._sanity_check
-        variables = {var.prefixed_name: var for var in self.inputs + self.outputs if var.units is not None}
+            # Any model variable that is redefined by an equation is treated as an input
+            if not pvar.local:
+                pvar.update(input=True)
 
-        for eq in self.equations:
+        # Resolve references to model variables
+        # If multiple references point to the same model variable, merge them
```
jonc125, line 1364: We should keep a record of the references though so they appear as distinct inputs/outputs. Though I don't think we have any instances of that in models!

fc/parsing/actions.py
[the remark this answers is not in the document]
jonc125: OK

fc/parsing/actions.py
```diff
@@ -1200,204 +1268,353 @@ def _number_generator(self, value, units):
...
+                continue
+            units = self.units.get_unit(pvar.units)
+
+            # Convert model_variable units
+            if (pvar.input or pvar.output) and pvar.model_variable is not None:
+                if pvar.model_variable.units != units:
+                    direction = DataDirectionFlow.INPUT if pvar.input else DataDirectionFlow.OUTPUT
+                    pvar.model_variable = self.model.convert_variable(pvar.model_variable, units, direction)
+
+            # Convert transitive output units
```
jonc125, line 1441: I need to think through whether it's possible to do this correctly with your new `_process_protocol_variables` approach, but it's late and my brain is tired. There were definitely edge cases that meant output conversion *had* to happen after all other processing before!

fc/parsing/actions.py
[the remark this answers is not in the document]
jonc125: True. OK!

fc/parsing/actions.py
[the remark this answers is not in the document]
jonc125: No, input here is a technical term that means specifically that the rest of the protocol (outside the model interface) is allowed to change the value of the variable at arbitrary time points. Which is why unit conversion for inputs needs to be handled differently: you need to ensure that when the protocol does this, the rest of the model sees the unit-converted value. Time *is* implicitly set from the protocol (via the solver) so does get converted using the input rule.

But arbitrary `define` clauses don't necessarily need or want this special processing. And except for ODEs or constants, it's actually inconsistent to have a `define` with an input on the LHS, because that results in an over-defined variable. States can change instantaneously, variables defined always to equal the result of a calculation can't. A define clause isn't changing the value of a variable at a particular time point, it's changing the always-true mathematical relationships between variables in the model.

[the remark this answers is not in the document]
MichaelClerx: OK, then I think I can just remove this line!

jonc125:
I've thought of a reasonably thorough way to test what is going on here, I think!

Suppose we create the following test model:
```
oxmeta:cytosolic_calcium_concentration units micromolar = 1
oxmeta:cytosolic_sodium_concentration units millimolar = 2
oxmeta:cytosolic_potassium_concentration units nanomolar = 3

oxmeta:extracellular_calcium_concentration units micromolar init 0
d(oxmeta:extracellular_calcium_concentration; oxmeta:time) = 1
oxmeta:extracellular_sodium_concentration units millimolar init 0
d(oxmeta:extracellular_sodium_concentration; oxmeta:time) = 2
oxmeta:extracellular_potassium_concentration units nanomolar init 0
d(oxmeta:extracellular_potassium_concentration; oxmeta:time) = 3

oxmeta:membrane_fast_sodium_current units microamps = 1
oxmeta:membrane_persistent_sodium_current units microamps = 2
oxmeta:membrane_background_sodium_current units microamps = 3
oxmeta:membrane_rapid_delayed_rectifier_potassium_current units microamps = 4

oxmeta:membrane_capacitance units microfarad = 10

oxmeta:membrane_voltage units millivolt init 0
d(oxmeta:membrane_voltage; oxmeta:time) = (oxmeta:membrane_fast_sodium_current + oxmeta:membrane_persistent_sodium_current + oxmeta:membrane_background_sodium_current + oxmeta:membrane_rapid_delayed_rectifier_potassium_current) / oxmeta:membrane_capacitance
```

And test with a protocol like this (units defs etc omitted for brevity):
```
model interface {
    output oxmeta:Concentration
    output oxmeta:ExtracellularConcentration units micromolar
    output oxmeta:CytosolicConcentration

    # This one we might prefer to raise a sensible error message rather than be legal!
    output oxmeta:extracellular_potassium_concentration units millimolar

    clamp oxmeta:cytosolic_calcium_concentration to 5000 nanomolar
    clamp oxmeta:cytosolic_sodium_concentration

    output oxmeta:membrane_voltage units millivolt

    output oxmeta:membrane_fast_sodium_current
    output oxmeta:membrane_persistent_sodium_current units nanoamps
    output oxmeta:membrane_background_sodium_current
    output oxmeta:membrane_rapid_delayed_rectifier_potassium_current

    input oxmeta:membrane_rapid_delayed_rectifier_potassium_current units milliamps = 0.014
    define oxmeta:membrane_fast_sodium_current = 11 milliamps / 1000 dimensionless
    define oxmeta:membrane_persistent_sodium_current = 22000 nanoamps
}

# Simulate from t=0 to 1 millisecond in one step

post-processing {
    # TODO: will fill in assertions later this morning!
}
```

fc/parsing/actions.py
MichaelClerx: Wouldn't it be an error for two inputs to point to the same variable that way?

fc/code_generation.py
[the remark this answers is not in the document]
MichaelClerx: Or we could update VariableDummys so you _can_ get the model. They're tied to a model anyway, in a non-reusable manner.

I'm thinking of ProtocolVariable as a "dumb" container for whatever data you put in it, so unless it's something you can read off of its (current) properties I'm a bit hesitant to put a method in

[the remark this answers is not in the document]
jonc125: That would be fine by me; perhaps as a weak reference so it doesn't cause potential issues for garbage collection with cyclic references?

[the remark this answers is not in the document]
MichaelClerx: I've never used them! I thought Python was OK at detecting that kind of cycle?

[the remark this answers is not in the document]
jonc125: That's true; I just looked up the weakref docs and it's more targeted at caches, which is a different case. (I'd come across them in some Django debugging yesterday which is what triggered my thought!)

fc/parsing/actions.py
jonc125: It would kind-of indicate an inconsistency in the ontologies used, since a variable shouldn't is-a two different concepts in the way we've set up the oxmeta ontology. But this is possible in other ontologies since they consider things slightly differently.

fc/code_generation.py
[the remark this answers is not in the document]
MichaelClerx: Interesting!

fc/parsing/actions.py
MichaelClerx: Oh, because the states are added as transitive variables. Cool!

MichaelClerx: Cool!

fc/parsing/actions.py
MichaelClerx: But it could create a situation where I say "set x to 1", and then "set y to 2", not realising they both set `model.z` ?

MichaelClerx: So a warning at the very least, but I'm leaning towards an error, because a user probably doesn't want a simulation to run this way?

MichaelClerx:
Summarising as:

- [ ] Inputs: Keep track of all terms that refer to the same variable, possibly give an error if there's 2
- [ ] Outputs: Keep track of terms pointing to same var, create output for all in code gen
- [ ] "Categories": Call this vector outputs, check that a vector output is never also a scalar output
- [ ] Create and run the test above, look for units issues
- [ ] May need to save finding of transitive outputs until after model manipulation
    - But then how will I know that they're transitive terms and not raises errors when no variable is found?
    - If outputs are treated differently again, we could even consider storing `ProtocolVariable.outputs = {rdf_term_1: units_1, rdf_term_2: units_2}`, but this would require creating extra variables (same for anything included in a vector output)

fc/parsing/actions.py
jonc125: That would be reasonable in the first instance.

jonc125:
The way pycml handled vector outputs was to get_variables_transitively early on, and if it matched multiple variables, note that this was a vector output, but then don't do anything special until the end when it re-ran get_variables_transitively to see what was now picked up. That may not be the best approach now; hopefully the test above will clarify!

MichaelClerx:
I suppose you could have an optional variable that gets created, and _then_ matches a transitive term, so would have to be added at the end?

jonc125:
Good thought! Yes. We should add that example to our test above :)
________________________________

MichaelClerx:
> Outputs: Keep track of terms pointing to same var, create output for all in code gen

Don't think I can do this one until we've implemented #169 ? And I've never touched the code that uses the generated model code, so not sure what the ramifications would be?

jonc125:
> > Outputs: Keep track of terms pointing to same var, create output for all in code gen
> 
> Don't think I can do this one until we've implemented #169 ? And I've never touched the code that uses the generated model code, so not sure what the ramifications would be?

Nope, this is entirely separate, since it's just the map from ontology term to model variable that needs extending. So e.g. when generating `output_info` you could have multiple entries with the same `var_name` field (but different `local_name`).

MichaelClerx:
It's not about the `var_name` field, it's about the `local_name` field, which should be a `name` field

jonc125:
> It's not about the `var_name` field, it's about the `local_name` field, which should be a `name` field

Yep, so as said on #169, for now just assert that each `local_name` is distinct, and we can support overlaps in a separate PR.

MichaelClerx:
How should transitive variables be ordered? Alphabetically? Order of appearance?

jonc125:
> How should transitive variables be ordered? Alphabetically? Order of appearance?

For state variables we've used order_added (model order) so it matches the internal solver order. For other vectors pycml does
```python
vector_outputs.sort(key=lambda v: self.var_display_name(v))
```
so alphabetically. Docstring for the var_display_name method:

> If it has an oxmeta name, uses that.  Otherwise, looks first for a bqbiol:is annotation,
> or uses the cmeta:id if present, or the name attribute if not.  If there is an interface
> component, strip the name of it out of the display name.

MichaelClerx:
@jonc125 it still fails on the test that requires order-by-display-name, but probably safe to start a final review of the rest now

MichaelClerx:
Or not

fc/parsing/actions.py
```diff
@@ -1200,204 +1266,301 @@ def _number_generator(self, value, units):
...
+                        # TODO: There's probably a faster way to check if we can resolve all variables in the RHS
+                        try:
+                            rhs = rhs.to_sympy(self._variable_generator, self._number_generator)
+                        except MissingVariableError as e:
+                            # Unable to create at this time, but may be able to at a next pass
+                            error = e
+                            continue
+
+                        # Set units based on RHS
+                        units = self.units.evaluate_units(rhs)
```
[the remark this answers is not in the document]
MichaelClerx: This has come up again. Without having units here I can't create this variable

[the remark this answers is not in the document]
MichaelClerx: I wasn't thinking straight when I made this change :D

`set_lhs_units_from_rhs` is not something we can do in cellmlmanip, is it? Because the variable units are fixed when the variable is created?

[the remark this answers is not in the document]
MichaelClerx: So is the idea that I set them to e.g. dimensionless first, and then have the RHS override it? I think that would hide errors
Also the code in `set_lhs_units_from_rhs` doesn't deal with state variables, it seems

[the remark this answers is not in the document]
MichaelClerx: I don't think we should have a `set_lhs_units_from_rhs`, really. It changes a variable's units without updating the model equations.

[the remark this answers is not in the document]
MichaelClerx: I'll just deduce the correct units with `convert_expression_recursively` when I create the variable instead

fc/parsing/actions.py
```diff
@@ -754,15 +745,13 @@ class OptionalVariable(BaseGroupAction, VariableReference):
     """
     def __init__(self, s, loc, tokens):
         super().__init__(s, loc, tokens)
-        if 'default' in self.tokens:
-            self.default_expr = s[self.tokens['default_start']:self.tokens['default_end']]
-        else:
-            self.default_expr = ''
+
+        self.default_expr = self.tokens.get('default', None)
```
jonc125, line 749: No `_expr` method needed?

fc/parsing/actions.py
```diff
@@ -790,9 +779,11 @@ def _expr(self):
         assert 1 <= len(self.tokens) <= 2
         name = self.tokens[0]
         if len(self.tokens) == 1:
+            # Clamp to initial value
```
jonc125, line 782: Perhaps add ', whatever that happens to evaluate as'?

fc/code_generation.py
```diff
@@ -195,42 +191,68 @@ def derivative_name(deriv):
...
+                pass
+                # TODO: Add some default ordering for consistent output?
             length = len(variables)
             var_name = [{'index': i, 'var_name': variable_name(s)} for i, s in enumerate(variables)]
+            output_variables.update(variables)
+        else:
+            continue
+
+        # TODO: Add an output for each rdf term pointing to the same variable.
+
```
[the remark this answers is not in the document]
jonc125: But not in this PR!

fc/parsing/actions.py
```diff
@@ -882,6 +876,178 @@ def _expr(self):
...
+
+    ``name``
+        A prefixed ontology term or a local variable name.
+    ``short_name``
+        A short name (no prefixes) that can be used for this variable.
+    ``long_name``
+        This variable's name plus any known aliases.
+    ``input_terms``
+        A list of rdf terms by which this variable is known as an input. This is typically zero or one term, but could
+        end up being more if two ontology terms specify the same model variables. For inputs this is an error.
```
jonc125, line 893: Actually, I don't think this does _need_ to be an error for inputs either, provided units are consistent?

fc/parsing/actions.py
```diff
@@ -882,6 +876,178 @@ def _expr(self):
...
+    ``output_terms``
+        A list of rdf terms by which this variable is known as a scalar output. This is typically zero or one term, but
+        multiple ontology terms may annotate the same output (as long as their units are consistent).
+    ``vector_output_terms``
+        A list of rdf terms by which this variable is known as a vector output. This is typically zero or one term.
+    ``is_input``
+        ``True`` iff the associated model variable will be set as a protocol input.
```
jonc125, line 900: 'set as a model input by the protocol' might be clearer language?

fc/parsing/actions.py
```diff
@@ -882,6 +876,178 @@ def _expr(self):
...
+    ``is_output``
+        ``True`` iff the associated model variable(s) will be read as a protocol output.
```
jonc125, line 902: Similarly 'read as a model output by the protocol'? Since the protocol itself also has inputs and outputs, which are different things!

fc/parsing/actions.py
```diff
@@ -882,6 +876,178 @@ def _expr(self):
...
+
+    def update(self, name=None, input_term=None, output_term=None, is_optional=False, is_vector=False, is_local=False,
+               units=None, initial_value=None, default_expr=None, equation=None, model_variable=None,
+               vector_variables=None):
+        """
+        Merges new information into this :class:`ProtocolVariable`, raises a ProtocolError if new information is
+        incompatible with what is already stored.
+        """
+        # Add name
+        # TODO: Determine this based on input and output terms?
```
jonc125, line 953: Not possible for local vars though

fc/parsing/actions.py
```diff
@@ -882,6 +876,178 @@ def _expr(self):
...
+        else:
+            assert not self.vector_variables, f'{self.long_name} is not a vector, but has vector variables'
+
+    def merge(self, pvar):
+        """
+        Merges the information from another protocol variable (``pvar``) into this one, assuming they both represent the
+        same model variable: raises a ProtocolError if any conflicts are found.
+        """
+        if self.model_variable is None or self.model_variable is not pvar.model_variable:
+            raise RuntimeError(
```
jonc125, line 1023: `ValueError` maybe?

test/models/transitive_variables.cellml
```diff
@@ -0,0 +1,194 @@
...
+  <units name="mM_per_ms">
+    <unit units="metre" exponent="-3" multiplier="1000"/>
+    <unit units="second" exponent="-1"/>
+    <unit units="mole"/>
+  </units>
+  <units name="mM_times_1e_minus_6">
+    <unit units="metre" exponent="-3" multiplier="1e-06"/>
+    <unit units="mole"/>
+  </units>
+  <units name="mV">
```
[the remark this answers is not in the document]
MichaelClerx: Nothing. But it's a generated file and it was easier to just output a reduction in the 7 base units rather than have it try to do something fancy

fc/parsing/actions.py
MichaelClerx: Seems to work fine! (Once I deleted all old caches)
Will add in if I get errors later

fc/code_generation.py
```diff
@@ -195,42 +191,68 @@ def derivative_name(deriv):
...
+                order = vector_orderings[rdf_term]
                 variables.sort(key=lambda s: order[s.rdf_identity])
+            else:
+                pass
+                # TODO: Add some default ordering for consistent output?
             length = len(variables)
             var_name = [{'index': i, 'var_name': variable_name(s)} for i, s in enumerate(variables)]
+            output_variables.update(variables)
+        else:
+            continue
```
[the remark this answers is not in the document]
MichaelClerx: There's a vector_ordering for the states, but I don't see where any of the other vector outputs define an order?

fc/code_generation.py
```diff
@@ -255,6 +277,43 @@ def derivative_name(deriv):
             'parameter_index': parameter_variables.get(eq.lhs, None),
         })
 
+    # Debug output
```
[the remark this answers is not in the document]
MichaelClerx: Could do, but wouldn't know how to see it when debugging then?

fc/parsing/actions.py
jonc125: Yes, it would get confusing! I'm happy to make it an error.

fc/code_generation.py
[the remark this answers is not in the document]
jonc125: Indeed, probably not worth doing as part of this PR, but maybe open an issue to implement it later?

fc/code_generation.py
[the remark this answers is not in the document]
jonc125: At present they don't, but the logic is there in `code_generation.py` so it can be supported when needed.

fc/code_generation.py
```diff
@@ -198,7 +198,7 @@ def derivative_name(deriv):
     for pvar in protocol_variables:
         if pvar.is_input and pvar.model_variable is not None:
             eq = model.get_definition(pvar.model_variable)
-            if eq is not None and not eq.lhs.is_Derivative and len(eq.rhs.atoms(VariableDummy)) == 0:
+            if eq is not None and not eq.lhs.is_Derivative and model.is_constant(pvar.model_variable):
```
jonc125: I meant to replace these lines with
```python
if pvar.is_input and pvar.model_variable is not None and model.is_constant(pvar.model_variable):
```
We don't need `eq` at all except for the constant check.

fc/parsing/actions.py
```diff
@@ -1532,8 +1531,13 @@ def _process_protocol_variables(self):
...
 
-                        # Units unknown! Will be set later based on RHS.
-                        units = None
+                        # Get units for the variable to create from its RHS, fixing inconsistencies if required
+                        units, rhs = self.units.evaluate_units_and_fix(rhs)
+
+                        # If an LHS is provided, the new variable could be a state variable, and so we'll need to
+                        # multiply the RHS units by the time units to get the variable units.
+                        if pvar.equation is not None and pvar.equation.is_ode:
+                            units *= self.time_variable.units
```
jonc125, line 1540: This looks good to me.

fc/code_generation.py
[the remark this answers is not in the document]
MichaelClerx: I'm updating it now to store a vector_ordering for all. But also the `variables` argument above is defined as `variables = list(pvar.vector_variables)` so maybe we can get rid of `vector_orderings` and just make sure the list `vector_variables` has the order we want?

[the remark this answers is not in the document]
MichaelClerx: (It definitely works, I'm just asking if there's any reason we need to stick to `vector_orderings`)

MichaelClerx:
Possible simplification now is dropping `vector_orderings`. Other than that I think we're done?

MichaelClerx:
Currently failing tests because I'm expecting a different thing to happen when we call `get_display_name()` without specifying an ontology. Can adapt once it's clear what the desired behaviour is :-)

fc/code_generation.py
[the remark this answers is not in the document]
jonc125: We needed it originally because code_generation was calling get_transitive... which didn't guarantee ordering. If we're now passing it a list anyway then yes we can just order the list.

MichaelClerx:
Ready for final review @jonc125

fc/parsing/actions.py
```diff
@@ -1052,111 +1220,34 @@ def modify_model(self, model, units):
...
         # Process ``clamp`` statements with an RHS
         self._handle_clamping()
 
-        self._annotate_state_variables()
+        # Process vector outputs
+        self._process_transitive_variables()
 
-        # Convert units for output variables
-        output_variables = self._convert_output_units()
+        # Gather state variables, store vector ordering and add to ProtocolVariable if requested
```
jonc125, line 1244:
```suggestion
        # Gather state variables, store vector ordering and add to protocol variables if requested
```

fc/parsing/actions.py
```diff
@@ -1200,204 +1294,341 @@ def _number_generator(self, value, units):
...
+            # Merge, and mark pvar for removal
+            partner.merge(pvar)
+            aliases.add(pvar)
 
-            # Variable names must be unique (can't even be re-used in imported/nested protocols)
-            if var.name in self.local_vars:
-                raise ProtocolError(f'Variable "{var.name}" was defined by more than one var statement.')
+        # Remove 'alias' references
+        for pvar in aliases:
+            del name_to_pvar[pvar.name]
```
jonc125, line 1401:
```suggestion
            del name_to_pvar[pvar.name]
        del aliases
```
To make clear that these are not needed in the rest of the method, but `var_to_pvar` is.

fc/parsing/actions.py
```diff
@@ -1200,204 +1294,341 @@ def _number_generator(self, value, units):
...
 
-            # Create and store variable
-            name = self.model.get_unique_name('protocol__' + var.name)
-            units = self.units.get_unit(var.units)
-            self.local_vars[var.name] = self.model.add_variable(name, units, initial_value=var.initial_value)
+        # Raise error if a variable is specified as an input using two different ontology terms
+        for pvar in name_to_pvar.values():
+            if len(pvar.input_terms) > 1:
+                raise ProtocolVariable(
+                    f'The mode variable {pvar.model_variable} is specified as a protocol input by more than one'
```
jonc125, line 1407:
```suggestion
                    f'The model variable {pvar.model_variable} is specified as a protocol input by more than one'
```

MichaelClerx:
@jonc125 optional outputs?

jonc125:
> @jonc125 optional outputs?

I thought they would have been removed by this point? Maybe not. No tests failed though!

MichaelClerx:
I think the protocol_variables list just gives all the information about variables specified in the protocol, including optional ones that couldn't be resolved. So probably if we add a test with an optional output that's also optional in post-processing this could fail?

jonc125:
Quite possibly. In which case please do revert that change! (Perhaps with a comment explaining the situation.)

fc/parsing/actions.py
MichaelClerx: Nope! The magic ProtocolVariable is already in the list of protocol variables. This is about adding the list of state variables to the magic ProtocolVariable

jonc125: Ah, I see!

MichaelClerx:
I agree it's a very ugly if-block :sweat_smile:

MichaelClerx:
Fixed now?

jonc125:
Huh, except that the assertion keeps tripping?

MichaelClerx:
I forgot to set `is_vector` on the magic state_variable annotation. Fixed now!

PR #159, opened 2020-04-15
Add unit conversion for new/changed equations

### LINKED ISSUE #106: Support simple protocol imports and units_of()

### LINKED ISSUE #107: Merge model interface from nested protocol

jonc125:
Will fix #106 by addressing the units conversion aspect.

Also fixes #107.

Requires https://github.com/ModellingWebLab/cellmlmanip/pull/265; `.travis.yml` will need changing back once that is merged.

### DIFF
--- a/fc/protocol.py
+++ b/fc/protocol.py
@@ -239,9 +239,9 @@ class Protocol(object):
         self.input_env.execute_statements(self.inputs)
 
         # Store unit definitions and add these to protocol
-        self.unit_definitions = {udef.name: udef for udef in details.get('units', [])}
-        for name, udef in self.unit_definitions.items():
-            self.units.add_unit(name, udef.pint_expression)
+        for udef in details.get('units', []):
+            self.unit_definitions[udef.name] = udef
+            self.units.add_unit(udef.name, udef.pint_expression)
 
         def merge_unit_definitions(unit_definitions):
             """Merge ``unit_definitions`` from a nested/imported protocol with ours.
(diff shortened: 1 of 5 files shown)


fc/protocol.py
```diff
@@ -241,9 +239,31 @@ def _parse(self):
         self.input_env.execute_statements(self.inputs)
 
         # Store unit definitions and add these to protocol
-        self.unit_definitions = details.get('units', [])
-        for units in self.unit_definitions:
-            self.units.add_unit(units.name, units.pint_expression)
+        self.unit_definitions = {udef.name: udef for udef in details.get('units', [])}
```
MichaelClerx, line 242: Would it make more sense to just to `self.unit_definitions = {}` and then iterate only once?

PR #167, opened 2020-04-22
Run test_other_ontologies

MichaelClerx:
See #141


MichaelClerx:
Following on from #158:

I think the bit with the output variables can be tidied up further:

- make `get_variables_transitively` return variables + terms, apply this to outputs in `collate` to create ProtocolVariables for these outputs
- stop calling `get_variables_transitively` in `_convert_output_units` (now becomes a simple loop)
- store outputs better, and use this in protocol.py to pass to code generation, stop calling `get_variables_transitively` in `code_generation.py`
- presumably the rule is "If a non-optional output doesn't resolve directly, but does resolve transitively, then that's fine."?

MichaelClerx:
> make get_variables_transitively return variables + terms, apply this to outputs in collate to create ProtocolVariables for these outputs

The rdf term in a protocol variable is used to uniquely identify the variable, which wouldn't be possible here as `get_variables_transitively` also searches using `bqbiol:isVersionOf`

MichaelClerx:
Continued in #158

PR #173, opened 2020-04-27
Check that units_of is evaluated at the right time

MichaelClerx:
See #106 

If units get redefined, it's important that units_of points to the new units.

Units can get set in two ways:

1. through an input/output statement
2. when a new variable is created

Thoughts so far:

- `units_of()` can only occur in the units of a number
- the `units_of` is evaluated when the number is converted to sympy
- the expressions don't get converted until existing variables have been converted. So case 1 is solved.
- I've updated the code so that if a number can't find the variable for its `units_of()` it raises a MissingVariableError.
- This then propagates and is caught in the model modification code (which will stop processing and hope to fix it on the next pass). So `units_of` is taken into account when creating and modifying variables. This solves case 2.

Still need to add tests

### DIFF
--- a/test/protocols/test_units_of.txt
+++ b/test/protocols/test_units_of.txt
@@ -1,0 +1,71 @@
+#
+# Tests using units_of
+#
+# Designed to run with simple_ode.cellml, which contains only three annotations:
+#  - cytosolic_sodium_concentration
+#  - membrane_voltage
+#  - time
+#
+
+namespace oxmeta = "https://chaste.comlab.ox.ac.uk/cellml/ns/oxford-metadata#"
+
+units {
+    molar = mole . litre^-1
+    millimolar = milli molar
+    ms = milli second
+    MV = mega volt
+}
+
+model interface {
+    independent var units ms
+
+    # Don't redefine units of cytosolic_sodium_concentration
+    input oxmeta:cytosolic_sodium_concentration = 1
+    output oxmeta:cytosolic_sodium_concentration
+
+    # Create a variable in the units of yet-to-be created variables (mM)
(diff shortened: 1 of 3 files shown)


MichaelClerx:
Have done this as a PR on #158 hoping that that'll get merged so that this becomes a PR on master

test/protocols/test_units_of.txt
```diff
@@ -0,0 +1,71 @@
...
+
+    # Create a variable in converted units of existing variable
+    define oxmeta:sodium_reversal_potential = 1 :: units_of(oxmeta:membrane_voltage)
+    output oxmeta:sodium_reversal_potential
+
+    # Convert units of membrane voltage to mega volts
+    input oxmeta:membrane_voltage = 0.1
+    output oxmeta:membrane_voltage units MV
+
+    # Create another variable in mV, to read out the value
```
jonc125, line 50:
```suggestion
    # Create another variable in volts, to read out the value
```

test/test_units_of.py
```diff
@@ -0,0 +1,16 @@
+"""
+Tests the units_of() construct when giving units to numbers.
+"""
+import fc
+
+
+def test_unit_of():
```
jonc125, line 7:
```suggestion
def test_units_of():
```

test/test_units_of.py
```diff
@@ -0,0 +1,16 @@
...
+    # Tests the units_of construct
+
+    proto_file = 'test/protocols/test_units_of.txt'
+    proto = fc.Protocol(proto_file)
+    proto.set_output_folder('test_unit_of')
```
jonc125, line 12:
```suggestion
    proto.set_output_folder('test_units_of')
```

PR #174, opened 2020-04-27
Run test_other_ontologies

MichaelClerx:
See #141 

Have done this as a PR on #173  hoping that that'll get merged so that this becomes a PR on master


MichaelClerx:
This uses the syntax:
```
    initial_V = sim1:test:V
    initial_param = sim2:test:input
```
whereas we'd normally cut off the prefix, and write e.g. ``sim1:ExtracellularConcentration``, as happens further down in the same protocol.

Not sure what the intended behaviour here is @jonc125 !

jonc125:
This syntax should be supported already by code outside the model interface, accessing the current state of those variables via the `ModelWrapperEnvironment`.

MichaelClerx:
Could you explain that syntax to me? I don't get it. Elsewhere when we refer to variables in another ontology we still just write `sim1:var_from_ontology_test`, not `sim1:test:var_from_ontology_test`. What makes this different?

jonc125:
See how name resolution works at https://chaste.cs.ox.ac.uk/trac/attachment/wiki/FunctionalCuration/Name%20resolution.png

MichaelClerx:
OK. Could you have a look at the travis test result? I have no idea what this fails. Using variables from other ontologies in post-processing works fine in other places. So maybe the issue is with using it in the library section?

jonc125:
I _think_ what is happening is that the `initial_V` reference is returning the entire state vector array rather than a single value, and so is changing as the simulation progresses. Which is definitely a bug somewhere, just not sure where yet :)

jonc125:
Happy to merge if you are @MichaelClerx !

MichaelClerx:
That's 3 today :D

PR #177, opened 2020-04-28
Clamp optional variable to initial value

MichaelClerx:
Currently not handled well

@jonc125 think it might be easier to integrate this type of clamps into the ProtocolVariable system?


MichaelClerx:
Is a "clamp" (without a "to") always meant to happen after all other modifications?

MichaelClerx:
(Not sure that matters too much? Since its new definition is just a constant. Might have systems that are unresolvable while it's more than just a constant?)

MichaelClerx:
Or are we waiting for all modifications to have occurred before we evaluate what its initial value is? (in case it's not a state)

And if so, does that mean we _should_ be allowed to first redefine it, and then clamp it?

jonc125:
It probably does make sense to integrate it to avoid confusion / being missed out.

The typical use-case will be clamping to an initial value given in the model, which might be a constant or computed as a constant. I don't think we allow any variable references on the RHS, and we don't allow a `define` with the same LHS, so it should be possible to resolve in any order?

test/protocols/test_clamping_optional.txt
```diff
@@ -0,0 +1,37 @@
...
+    mM = milli mole . litre^-1
+}
+
+model interface {
+    independent var units ms
+
+    output oxmeta:membrane_voltage units mV
+    clamp oxmeta:membrane_voltage
+
+    output oxmeta:extracellular_chloride_concentration units mM
```
jonc125, line 19: There's no optional statement here?

test/protocols/test_clamping_optional.txt
```diff
@@ -0,0 +1,40 @@
...
+units {
+    mV = milli volt
+    ms = milli second
+    mM = milli mole . litre^-1
+}
+
+model interface {
+    independent var units ms
+
+    # Clamp a variable that's know to exist
```
jonc125, line 16:
```suggestion
    # Clamp a variable that's known to exist
```

Issue #156, opened 2020-04-09
Unit conversion rules: add tests and support


MichaelClerx:
Choosing rule is easy*:

- if old unit not compatible with new unit
  - scan for rule that has input comp with old, output compat with new

*provided we never have to apply more than one rule

Applying:

Not sure yet!

MichaelClerx:
Look in e.g. ICaL.txt for examples

MichaelClerx:
```
    convert A_per_F to uA_per_cm2 by lambda rhs: rhs * chaste_membrane_capacitance
    convert uA_per_cm2 to A_per_F by lambda rhs: rhs / chaste_membrane_capacitance
    convert microamps to uA_per_cm2 by lambda rhs: (rhs / oxmeta:membrane_capacitance) * chaste_membrane_capacitance
    convert uA_per_cm2 to microamps by lambda rhs: (rhs * oxmeta:membrane_capacitance) / chaste_membrane_capacitance
```

MichaelClerx:
The rule for inputs is:
```
x = 5mV ox:x
y = 4 + x
```
where `ox:x` denotes the oxmeta annotation `x`
becomes
```
x2 = f(5mV)  ox:x
x = f'(x2)
y = 4 + x
```
so we need both f and f' here, the documentation isn't explicit about that!

For outputs it's easier:
```
x = y ox:x
```
becomes
```
x2 = f(x) ox:x
x = y
```
but what about the time variable?

MichaelClerx:
https://pint.readthedocs.io/en/0.11/contexts.html#defining-contexts-programmatically

jonc125:
The documentation isn't particularly explicit because it 'just happens' as part of the general units conversion functionality in pycml. Rather than throwing an error if at any point you're asked to convert between dimensions, it checks if it has a rule that lets you do it. If you do, it applies the rule, and now you're back in the scaling-within-a-dimension case that it knows how to handle.

I think the pint context approach should deliver the same functionality, judging from the docs!

jonc125:
Also see developer ref docs at https://pint.readthedocs.io/en/0.11/developers_reference.html#pint.Context

jonc125:
The following works with the small patch to `units.py` shown below:
```python
import pint
import sympy as sp
from cellmlmanip.model import VariableDummy
from cellmlmanip.model import UnitStore

store = UnitStore()
microampere = store.add_unit('microampere', '1e-6 * ampere')
[...]

Issue #12, opened 2019-01-25
Update logging code to use the `logging` module


MichaelClerx:
See e.g. `protocol.log_progress`, `protocol.log_warning`

MichaelClerx:
Regarding warnings, https://docs.python.org/3/howto/logging.html#when-to-use-logging does suggest that `log_warning` is the way to go here

MichaelClerx:
Closing this, don't think it's useful anymore

PR #179, opened 2020-04-30
Unit conversion rules

MichaelClerx:
See #156

### DIFF
--- a/fc/parsing/actions.py
+++ b/fc/parsing/actions.py
@@ -1566,7 +1566,7 @@ class ModelInterface(BaseGroupAction):
 
             # Convert expression to sympy
             try:
-                expr = expr.to_sympy(lambda x: variable_generator(x, free), self._number_generator)
+                expr = expr.to_sympy(lambda x: variable_generator(x, free), number_generator)
             except ProtocolError as e:
                 print('  _[]_')
                 print('W (")')
@@ -1576,7 +1576,7 @@ class ModelInterface(BaseGroupAction):
                 continue
 
             # Add transformation to pint context
-            context.add_transformation(u1, u2, lambda ureg, rhs: expr.xreplace({UNIT_LAMBDA_SYMBOL: rhs}).evalf())
+            context.add_transformation(u1, u2, lambda ureg, rhs: expr.subs({UNIT_LAMBDA_SYMBOL: rhs}).evalf())
 
         # Store and enable context
         self.units._registry.add_context(context)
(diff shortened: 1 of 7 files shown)


MichaelClerx:
Hi @jonc125 !

Is this the kind of approach you had in mind?
I'm assuming we want all the numbers and variables as pint quantities, but not sure that plays well with sympy?

fc/parsing/actions.py
```diff
@@ -1452,6 +1474,128 @@ def get(ref):
...
+
+            # Convert expression to sympy
+            try:
+                expr = expr.to_sympy(lambda x: variable_generator(x, free), number_generator)
+            except ProtocolError as e:
+                print('  _[]_')
+                print('W (")')
+                print('|-(:\')-<')
+                print('|(\'  )')
+                print('JON HOW DO I RAISE A WARNING? ' + e.short_message)
```
jonc125, line 1574: pycml used to just print to stderr. Using the logging module would be a nicer approach!

jonc125:
Which come to think of it might mean we need a way to convert 
`simple_expr` to a Python expression rather than Sympy to do both the 
pint unit arithmetic and Sympy algebra together? Or maybe create a 
restricted version of that for conversion rules, since I can't see rules 
actually needing anything beyond plus/minus/times/divide/pow as operators?

jonc125:
Thinking further, the `to_sympy` method for those operators could just use Python operations, then we could reuse that - the Python operations would result in Sympy expressions if applied to Sympy atoms/symbols.

MichaelClerx:
> Not had a chance to properly look at this. But I was hoping we could make the conversion rule behave such as to provide a get_conversion_factor result that would have magnitude a sympy expression implementing the rule, with units containing the pint Unit version of the rule, or something like that?

Should we maybe work towards some new syntax where you just provide a factor that can be used both ways instead of two lambdas with the hidden assumption that they can be reduced to multiplication by some factor?

jonc125:
> Should we maybe work towards some new syntax where you just provide a factor that can be used both ways instead of two lambdas with the hidden assumption that they can be reduced to multiplication by some factor?

All the examples in the pint docs match that assumption. I'm not sure if you'd never want to raise to a power or take a root though? Or maybe even convert to a log scale? But such things might be better modelled explicitly rather than using a context/conversion rule?

fc/parsing/actions.py
MichaelClerx: OK someone can do that as part of #12 then

MichaelClerx:
> I'm not entirely convinced by reducing all conversion rules to multiplicative factors (it wouldn't work for the pint spectroscopy example `[length] <-> [frequency]: speed_of_light / n / value` or other transformations requiring reciprocals) but it's reasonable for a first implementation I think since it covers our existing use cases.

As far as I can tell `cellmlmanip` currently only supports multiplicative factors

MichaelClerx:
I don't like it either, because it's not consistent with the interface we present to the user, but agreed should be OK for now

Issue #109, opened 2020-02-23
Implement interpolate() support in the model interface


jonc125:
This is tested in `test_clamping.py:test_interpolation_clamp`.

Part of #23. Blocked by ~#97~ and probably ~#99~.

PR #183, opened 2020-05-03
Clamping to data file and interpolation clamp

MichaelClerx:
See #109


MichaelClerx:
Basic data clamp works, the `interpolate` construct doesn't.

PR #202, opened 2020-05-06
Slight tidying of test files

MichaelClerx:
All live in `test` now, but still separating the real life use cases from the models and protocols created just for testing


MichaelClerx:
@jonc125 what do you think of this proposal for tidying up the test files?
There's still a lot of unused files we could weed out at some later time

jonc125:
Oh, actually, some of the data files are *not* output so we need an `input` subfolder too. The interpolation_test CSV files, for instance. Those are the only ones I can spot at a quick look though.

The `*divisions.dat` files can be removed, they were only used by cell-based Chaste C++ tests. Similarly you can remove:
- reference_traces
- sicouri_dog_ventricle
- sun_rat_ventricle
- morgan_human_ventricle
- kim_human_atrial

MichaelClerx:
> I've not gone through every change, but this looks reasonable and the tests pass. I'd probably have gone for a `models/real` structure rather than `real/models` but that's a minor point and subject to personal preference! What I would like to see before this is merged is a readme.md file added to the test folder outlining the structure and rationale.

Fair point! I initially had it as `suite/[models|protocols|data]`, reasoning that this was a suite of real life test cases you could run through e.g. python_reproducibility, but then I saw we had a hard-coded selection and actually use the real models/protocols a lot in other tests...

PR #210, opened 2020-05-07
Writing debug output from model generation to file in protocol output dir

### LINKED ISSUE #172: Write model info to file before code generation
So that users can use it for debugging.

At the moment we just print it to screen

MichaelClerx:
Closes #172.


fc/protocol.py
```diff
@@ -638,6 +638,7 @@ def set_model(self, model):
             # Create weblab model at path
             create_weblab_model(
                 path,
+                self.output_folder or temp_dir,
```
jonc125, line 641: Lots of tests are failing because the method expects a string but `self.output_folder` is an `OutputFolder` instance.

PR #211, opened 2020-05-07
Replace print statements by calls to logging module

MichaelClerx:
See #12

### DIFF
--- a/fc/protocol.py
+++ b/fc/protocol.py
@@ -80,9 +80,6 @@ class Protocol(object):
 
         The protocol must be specified using the textual syntax, as defined by the CompactSyntaxParser module.
         """
-        # Logging
-        self.log = logging.getLogger(__name__)
-
         # if the filename passed as argument to CompactSyntaxParser is not found
         # it tries to read it before checking it exists
         # and you get a messy exit
@@ -98,7 +95,8 @@ class Protocol(object):
         self.proto_file = proto_file
         self.proto_name = os.path.basename(self.proto_file)
 
-        self.log.info(f'Constructing {self.proto_name}.')
+        log = logging.getLogger(__name__)
+        log.info(f'Constructing {self.proto_name}.')
 
         # Nesting level: Zero for main protocol
         self.nesting_level = 0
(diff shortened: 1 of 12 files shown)


MichaelClerx:
Not sure what's going on here. Unable to reproduce locally

MichaelClerx:
That seems to have fixed it

fc/protocol.py
```diff
@@ -660,24 +670,3 @@ def get_path(self, base_path, path):
             new_path = os.path.join(library, path)
         return new_path
 
-    def set_indent_level(self, indent_level):
-        """Set the level of indentation to use for progress output."""
-        self.indent_level = indent_level
-
-    def log_progress(self, *args):
```
jonc125, line 667: In getting rid of these helper methods, you're no longer indenting the log output from nested protocols, which makes the output harder to follow. Could we recover that functionality, or something equivalent?

fc/templates/weblab_model.pyx
```diff
@@ -312,5 +299,5 @@ cdef class {{ class_name }}(CvodeSolver):
         """
         # TODO Update this (and rest of fc) to Python3
         # TODO Use logging here, or raise an exception
-        print >>sys.stderr, '  ' * self.indent_level, 'set_solver: Models implemented using Cython contain a built-in ODE solver, so ignoring setting.'
+        print >>sys.stderr, 'set_solver: Models implemented using Cython contain a built-in ODE solver, so ignoring setting.'
```
jonc125, line 302: Is there a reason for not using logging here?

test/test_simple_protocols.py
```diff
@@ -6,21 +6,21 @@
...
 
-def test_core_post_proc_txt():
+def test_core_postproc():
     proto_file = 'test/protocols/test_core_postproc.txt'
     proto = fc.Protocol(proto_file)
     proto.run()
 
 
-def test_graph_txt():
+def test_graph_state():
```
jonc125, line 21: This duplicates a test name in test_code_generation.py

fc/templates/weblab_model.pyx
MichaelClerx: Just that I'm not sure how to do it in cython

fc/protocol.py
MichaelClerx: I'm not sure!

Happy to drop this PR and issue if we prefer the old way.

I'm guessing you could write a better log handler in the `fc-runner` that splits the logged messages into files according to their source

jonc125: Problem is you've lost the info by the time it gets to the fc-runner.

I was anticipating keeping the `log_progress` etc methods but using the logging library inside them?

MichaelClerx: Yeah I thought about that but then I'm not sure we're using the logging framework in a sensible way.

As far as I understand the idea is that log everything to some central point, with both a logging level and some identifier for the messages' origin. Whatever happens with the logged messages, including the formatting, is then the responsibility of a handler.

So if we're going to impose our own structure / formatting on top of that, I don't really see the point in using it in the first place

MichaelClerx:
Closing this, don't think it's useful anymore

PR #212, opened 2020-05-18
Implement interpolation clamps

### LINKED ISSUE #109: Implement interpolate() support in the model interface

jonc125:
Fixes #109.

This can be reviewed, but don't merge until https://github.com/ModellingWebLab/cellmlmanip/pull/287 is merged and we revert .travis.yml.


fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
     return env.get_template(path)
 
 
+class DataInterpolation(Quantity):
+    """Represents linear interpolation on data loaded from file, for code generation."""
+    _next_id = 0  # Used to generate unique table IDs in code
+
+    # Sympy annoyingly overwrites __new__
+    def __new__(cls, name, data, index_variable, *args, **kwargs):
+        print('interp__new__', name)
```
MichaelClerx, line 69: Debug comment?

fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
...
+
+        # Ensure we depend on the index variable in the model's graph
+        obj._args = [index_variable]
+
+        return obj
+
+    def __init__(self, name, data, index_variable, units):
+        """Create a new interpolation construct.
+
+        :param name: the data file base name
```
MichaelClerx, line 84: I think ``name`` is only used as an identifier here, so doesn't have to equal a file name?

fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
...
+        :param data: 2d numpy array containing the column-wise data
```
MichaelClerx, line 85: Time/value ? Or can it be vector values too?

fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
...
+        :param index_variable: the :class:`cellmlmanip.model.Variable` used to index the data
+        :param units: the units of the interpolated values (a :class:`~cellmlmanip.units.UnitStore.Unit`)
+        """
+        print('interp__init__', name, index_variable, units)
```
MichaelClerx, line 89: Another print

fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
...
+        super().__init__(name, units)
+        self.data = data[1]
+        self.index_variable = index_variable
+        self.initial_index = '%.17g' % data[0, 0]
```
MichaelClerx, line 93: What's going on here? String representations of the first and final data points?

fc/code_generation.py
```diff
@@ -59,6 +60,67 @@ def load_template(*name):
...
+        """Set the name used for our index variable in code.
+
+        :param variable_name_generator: function turning our index variable into its name in the code
+        """
+        self.index_name = variable_name_generator(self.index_variable)
+
+    @property
+    def lookup_call(self):
+        """The code for a function call looking up this interpolation table."""
+        assert self.index_name is not None
```
MichaelClerx, line 115: `RuntimeError` ?
Looks like misuse of the interface rather than a sanity check?

fc/parsing/actions.py
```diff
@@ -873,19 +874,80 @@ def lhs_to_sympy(self, variable_generator, number_generator):
...
+                    # Vertical jumps are allowed in voltage clamp protocols, but trying to interpolate gives you a
+                    # divide by zero. Just skip the jump part, and we'll interpolate correctly at either side.
+                    continue
+                x1 = number_generator(data[0, row], self.index_units)
+                x2 = number_generator(data[0, row + 1], self.index_units)
+                y1 = number_generator(data[1, row], self.data_units)
+                y2 = number_generator(data[1, row + 1], self.data_units)
+                cond = sympy.And(index >= x1, index <= x2)
+                case = y1 + ((y2 - y1) / (x2 - x1)) * (index - x1)
+                pieces.append((case, cond))
```
MichaelClerx, line 949: We should probably have something written in the FC manual about what happens to the left and right of the signal in the data file, and make sure this implementation does that?

fc/parsing/actions.py
```diff
@@ -873,19 +874,80 @@ def lhs_to_sympy(self, variable_generator, number_generator):
...
+                    f'The data file {data_file_name} used in an interpolate() has all zero table steps.')
+            # Unit-convert the indexing variable if required
+            data_index_units = number_generator(1.0, self.index_units).units
+            index = index.model.convert_variable(
+                index, data_index_units, DataDirectionFlow.OUTPUT, move_annotations=False)
+            # Result is a special Symbol that can generate interpolation code
+            data_units = number_generator(1.0, self.data_units).units
+            return DataInterpolation(data_file_name, data, index, data_units)
+        else:
+            # Iterate over data rows to construct a piecewise representation for the interpolation
```
MichaelClerx, line 936: Seems a bit hacky?
Maybe leave for now and add a ticket to add something more sensible in the future, e.g. a `DataInterpolation` variant that finds the correct 2 values to linearly interpolate via bisection, or that replaces the data entirely with a cubic spline?

fc/code_generation.py
jonc125: It's only code in the same module that uses this bit of the interface though, so felt more like an internal consistency check, and not worth adding a test of the error case for.

fc/parsing/actions.py
jonc125: See the special case mentioned for `define` at https://chaste.cs.ox.ac.uk/trac/wiki/FunctionalCuration/ProtocolSyntax#Modelinterface

fc/code_generation.py
jonc125: Yep, for insertion into generated code.

fc/code_generation.py
jonc125: It's also used as the identifier in the generated code, and the data file name is what is passed in from actions.py.

fc/code_generation.py
jonc125: Index/value; both are vectors. I'll add to the docstring.

fc/parsing/actions.py
jonc125: The piecewise is hacky? Much easier than more special-case code though. This is intended for small voltage step protocols. You don't tend to have large numbers of steps I don't think? Certainly not in the samples I've been given when creating this feature!

fc/code_generation.py
MichaelClerx: Yeah that's what's passed in, but it's never used as a file name, right?

fc/parsing/actions.py
MichaelClerx: I don't see where that specifies how extrapolation is handled?

jonc125: We're not extrapolating, only interpolating between the bounds of the piece condition.

MichaelClerx: So what happens if the simulation goes beyond the bounds specified in the data? And what should happen? I think we need to define that

jonc125: Extrapolation is an error. Updated docs to say this.

PR #214, opened 2020-09-17
Add utilities needed by WL front-end

jonc125:
These methods allow us to replace pycml for the protocol analysis tasks. See also https://github.com/ModellingWebLab/fc-runner/pull/10 and https://github.com/ModellingWebLab/deployment/pull/34.

They also replace the obsolete command-line tools from the old FunctionalCuration project.


MichaelClerx (approved):
General strategy looks good!

Issue #215, opened 2020-09-28
Write test using static model, protocol, data set


MichaelClerx:
Must do whole fitting process, but from static (hardcoded / local) input, and ending up with local data (so no WL connections yet)

PR #216, opened 2020-09-28
Fitting test (with static input and output)

MichaelClerx:
See #215


jonc125 (commented):
Is `test/models/beattie-2017-ikr.csv` really a model? Would it be better in `test/input`?

fc/protocol.py
```diff
@@ -540,102 +543,150 @@ def set_input(self, name, value_expr):
...
+        del sys.modules['model']
 
-            # Create output folder
-            if self.output_folder:
-                temp_dir = tempfile.mkdtemp(dir=self.output_folder.path)
-            else:
-                temp_dir = tempfile.mkdtemp()
+        # Benchmarking
+        self.timings['load model'] = (
+            self.timings.get('load model', 0.0) + (time.time() - start))
```
jonc125, line 658:
```suggestion
            self.timings.get('load model', 0.0) + (time.time() - start))
        
        self._set_model(model)
```

Issue #59, opened 2019-09-16
Remove references to CHASTE_TEST_OUPUT and CHASTE_NUM_PROCS


MichaelClerx:
Made in 

- [ ] test/test_output_folder .py
- [ ] test/test_python_reproducibility.py
- [ ] fc/file_handling.py

MichaelClerx:
`CHASTE_TEST_OUTPUT` is currently [set in deployment](https://github.com/ModellingWebLab/deployment/blob/master/roles/celery/templates/config.json.j2#L16), but no sign of `CHASTE_NUM_PROCS`.

Regarding `CHASTE_TEST_OUTPUT`, @jonc125 do we currently use this location for _all_ (non-test) output? If not then not sure we need this

jonc125:
It's used for non-test output too. Blame Chaste's historical baggage :)

jonc125:
CHASTE_NUM_PROCS was only ever used in FC tests though iirc.

MichaelClerx:
OK, so when we migrate to the FC non-chaste backend, we can rename TEST_OUTPUT?

PR #217, opened 2020-10-20
Reinstate reproducibility testing

jonc125:


fc/test_support.py
```diff
@@ -81,70 +82,37 @@ def within_any_tolerance(arr1, arr2, rel_tol=None, abs_tol=None):
...
             actual = proto.output_env.look_up(name)
         except KeyError:
-            if os.path.exists(data_file):
-                results_ok = False
-                if messages is not None:
-                    messages.append("Output %s not produced but reference result exists" % name)
-            elif results_ok:
-                results_ok = None  # Indicate expected failure
+            assert not os.path.exists(data_file), f"Output {name} not produced but reference result exists"
             continue  # Can't compare in this case
```
MichaelClerx, line 109: Why isn't this one an error too?

test/test_python_reproducibility.py
```diff
@@ -1,265 +1,99 @@
...
+    # If there are missing reference results then this is expected to fail
+    data_folder = f'test/output/real/{model_name}/{proto_name}'
+    for name in expected_outputs:
+        if not os.path.exists(os.path.join(data_folder, 'outputs_' + name + '.csv')):
+            request.applymarker(pytest.mark.xfail(reason=f'Missing reference result {name}', strict=True))
+            break
+
+    # Try to run the protocol
+    proto = fc.Protocol('test/protocols/real/%s.txt' % proto_name)
+    os.environ['CHASTE_TEST_OUTPUT'] = str(tmpdir)
```
MichaelClerx, line 80: Are we using `CHASTE_TEST_OUTPUT` anywhere outside of `fc` now? Because if so we could probably communicate it in some better way, e.g. a global variable `fc.TEST_OUTPUT`?

MichaelClerx (approved):

fc/test_support.py
jonc125: The absence of a reference result indicates that we're expecting the simulation to fail, or otherwise not produce the output (maybe it's optional for older models).

test/test_python_reproducibility.py
jonc125: That's a good idea. I think we already have an issue to get rid of the env var?

MichaelClerx: Yeah in https://github.com/ModellingWebLab/weblab-fc/issues/59. Linking it here for future ref!

PR #232, opened 2021-01-22
Moving to github actions for testing

MichaelClerx:


fc/code_generation.py
```diff
@@ -65,7 +65,9 @@ class DataInterpolation(Quantity):
     _next_id = 0  # Used to generate unique table IDs in code
 
     # Sympy annoyingly overwrites __new__
-    def __new__(cls, name, data, index_variable, *args, **kwargs):
+    def __new__(cls, name, data=None, index_variable=None, *args, **kwargs):
+        assert isinstance(name, str), str(name) +' '+ str(type(name))
```
MichaelClerx, line 69: Why this line?

MauriceHendrix: I was trying to get it to behave like a proper quantity (it couldn't cope with the extra positional arguments), but ultimately I couldn't get it to work, so reverted & adding a work-around instead.

fc/parsing/actions.py
```diff
@@ -87,7 +87,7 @@
     E.Floor: sympy.floor,
     E.Max: sympy.Max,
     E.Min: sympy.Min,
-    E.Rem: sympy.mod,
+    E.Rem: sympy.core.mod,
```
jonc125, line 90: What version of sympy was this introduced in? We should add sympy to our explicit list of dependencies (rather than via cellmlmanip) and specify the minimum that we support.

fc/code_generation.py
```diff
@@ -132,6 +133,13 @@ def __init__(self, symbol_function=None, derivative_function=None):
         # Deal with _exp function introduced to avoid simplification
         self._function_names['_exp'] = 'math.exp'
 
+    def doprint(self, expr):
```
jonc125, line 136: Admittedly I don't understand the sympy printer design, but this seems an expensive way to resolve the issue! Is the replace recursive throughout the whole expression, or just at a single level? In any case, shouldn't we be able to define a `_print_DataInterpolation` method in this class which will be used, according to https://docs.sympy.org/latest/modules/printing.html?

MauriceHendrix: the problem is that the super().do_print does a replace optimisation that causes wierd behaviour **BEFORE** you would get to any `_print_DataInterpolation`. This call causes DataInterpolation.__init__ to be called with just a cellmlmanip.model.Variable as parameter and I can't figure out what to do in that case. If I just spit it back or make a copy I get a strange error, which I don't get with either the current work around or if I knowck out the replace optimisation in the printer.

It probably would be possible to avoid the replace, but that doesn't solve the real issue in that doing a replace (even on things that aren't even in the expression) causes this srang behaviour.

```
model.c:6748:15: warning: '__pyx_f_5model_lookup_data_table_2' defined but not used [-Wunused-function]
 static double __pyx_f_5model_lookup_data_table_2(PyObject *__pyx_v_var_protocol__table_key_converted) {
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
model.c:6609:15: warning: '__pyx_f_5model_lookup_data_table_0' defined but not used [-Wunused-function]
 static double __pyx_f_5model_lookup_data_table_0(PyObject *__pyx_v_var_time_converted) {
               ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
model.c:6470:15: warning: '__pyx_f_5model_lookup_data_table_1' defined but not used [-Wunused-function]
 static double __pyx_f_5model_lookup_data_table_1(PyObject *__pyx_v_var_protocol__derived2_converted) {
```

In fact without that replace the whole `_print_DataInterpolation` isn't even necessary (it's was previously solved in the `variable_name` method in code_generation.py)

jonc125: The `Variable` argument is probably because of the line `obj._args = [index_variable]` in `__new__`, so something is trying to create a new `DataInterpolation` just from its `_args`. Adding more things to `_args` might well break some other part of sympy's machinery, I don't know!

Looking at the latest code for `Symbol` etc., maybe we need to define at least `__getnewargs__`. Or actually from [`Basic`](https://github.com/sympy/sympy/blob/9e8f62e059d83178c1d8a1e19acac5473bdbf1c1/sympy/core/basic.py#L117) maybe define our own `copy` to do the right thing? Or our own [`func`](https://github.com/sympy/sympy/blob/9e8f62e059d83178c1d8a1e19acac5473bdbf1c1/sympy/core/basic.py#L682).

MauriceHendrix: Yes adding more things to args breaks other things I did try that. 
I did try define a copy/deepcopy but it doesn't get called.

I'll look into basic/func.

jonc125 (commented):
Looking good! I think the sympy version in dependencies is the only thing to address now.

fc/parsing/actions.py
MauriceHendrix: actually @MichaelClerx  @jonc125 shouldn't this be `Mod` rather than `mod`?
```
>>> type(sympy.core.mod)
<class 'module'>
>>> type(sympy.Mod)
<class 'sympy.core.function.FunctionClass'>
>>> sympy.core.mod(25,6)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'module' object is not callable
>>> sympy.Mod(25,6)
1
>>>
```

jonc125:
Indeed, oops 😊

fc/parsing/actions.py
MichaelClerx: I don't know

jonc125 (approved):

